Issue #48, opened 2024-04-25
Design interface coupling for History


jamienoss:
The existing interface of having ``get`` and ``set`` funcs seems complete for what this is.

Possible API additions:
 * ``last()`` (or equivalent) but this depends on the data type being used under the hood as this may already have "last" functionality. However, this may still be desirable to directly expose to the API layer.

The main question is how this couples with other interfaces, how is it used? 

Questions:

* Add this to ``evolver.base``? (not doing so could introduce circular imports, we'll have to see...)
* Make attr of ``evolver.BaseInterface``?
* What is the interface? Is there a clean opportunity to bake this under the hood, i.e., in ``Device.set()`` & ``Device.get()``?
* Make this mostly a decorator pattern, e.g., see above point about ``Device`` set| get. I think this would work quite elegantly - that funcs needing to be recorded in the history, simply get decorated. Could add an arg to spec whether the input or output (or both) get recorded.
* ...

PR #51, opened 2024-04-30
Add History.history decorator

jamienoss:
Relates to #48

### DIFF
--- a/evolver/history.py
+++ b/evolver/history.py
@@ -1,4 +1,4 @@
-import time
+import datetime
 from abc import abstractmethod
 from collections import defaultdict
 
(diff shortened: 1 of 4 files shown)


evolver/history.py
```diff
@@ -10,13 +10,41 @@ class Config(BaseConfig):
...
+
+                # Track history.
+                cls.getHistory().set(name=name, data=ret)
+
+                return ret
+            return wrapper
+        return decorator
+
+    @classmethod
+    def getHistory(cls, name: str = None):
```
jamienoss, line 44: @amitschang We have options here, what's your pref:
- make ``history.history`` a class attribute similarly to ``logging.Logger.manager``
- singleton
- some additional registration func
- something else?

evolver/history.py
```diff
@@ -10,13 +10,41 @@ class Config(BaseConfig):
...
-    def put(self, name, data):
+    def set(self, name, data):
         pass
 
     @abstractmethod
     def get(self, query):
         pass
 
+    @classmethod
+    def track_history(cls, name: str = None):
```
jamienoss, line 21: We can always create a historyserver interface and this there. At present the two seem either a little entangled or we're missing an internal aspect to the interface to allow for the backend ``history`` to be pluggable.

jamienoss:
No team buy-in so closing.

PR #52, opened 2024-04-30
Allow overrides when calling ConfigDescriptor.create()

jamienoss:


jamienoss:
@amitschang Note that this still won't fix the issue I described about the evolver being passed to controllers as ``self`` from the evolver instance and that the devices don't live in ``Controller.Config`` - because this pattern will still validate and unrecognized args are ignored.

amitschang (approved):

PR #55, opened 2024-04-30
Pass devices to controllers from evolver

### LINKED ISSUE #54: Refactor Controller such that it doesn't explicitly require an evolver instance
Refactor Controller such that it ~takes a device dict rather than an evolver instance~ doesn't explicitly require an evolver instance

jamienoss:
Resolves #54 

~Actually, this totally doesn't work either as the design that @amitschang you came up with to name the devices explicitly in the config, is better, which means that this just doesn't go in the interface at all, but then neither should the ``evolver``. So this PR should really only rip that out rather than replace it.~

### DIFF
--- a/evolver/device.py
+++ b/evolver/device.py
@@ -55,6 +55,8 @@ class Evolver:
                 del(self.hardware[name])
                 del(self.last_read[name])
         self.controllers = []
+        # NOTE: This has to occur after the hardware/device drivers have been instantiated as they get passed to the
+        # controller.
         for controller in config.controllers:
             self.setup_controller(controller)
         if config.serial is not None:
@@ -74,8 +76,14 @@ class Evolver:
         self.hardware[name] = driver_config.driver(self, config, calibrator)
         self.last_read[name] = -1
 
-    def setup_controller(self, controller):
-        self.controllers.append(controller.driver_from_descriptor(self))
+    def setup_controller(self, controller: ControllerDescriptor):
+        config = controller.config.copy()
+        # Creat dict of required hardware instances.
+        hardware = {x: y for x, y in self.hardware.items() if x in config}
+        # Update controller config with actual hardware instances.
+        config.update(hardware)
+        # Instantiate controller and add to list.
+        self.controllers.append(controller.driver(**config))
 
     def get_hardware(self, name):
         return self.hardware[name]
--- a/evolver/controller/interface.py
+++ b/evolver/controller/interface.py
@@ -1,27 +1,13 @@
 from abc import ABC, abstractmethod
 
-from evolver.base import BaseConfig, ConfigDescriptor
+from evolver.base import BaseConfig
 
 
 class Controller(ABC):
-    """
-        Base interface class for all control implementations.
-
-        Attributes:
-            devices (:obj:`dict` of :obj:`Device`, :obj:`dict` of :obj:`ConfigDescriptor`): Dictionary of devices to
-                control.
-    """
+    """ Base interface class for all control implementations. """
 
     class Config(BaseConfig):
-        devices: dict
-
-    def __init__(self, devices: dict):
-        self.devices = devices
-
-        # Instantiate objects from config descriptors.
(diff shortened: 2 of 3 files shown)


evolver/device.py
```diff
@@ -74,8 +76,14 @@ def setup_driver(self, name, driver_config: HardwareDriverDescriptor):
...
-    def setup_controller(self, controller):
-        self.controllers.append(controller.driver_from_descriptor(self))
+    def setup_controller(self, controller: ControllerDescriptor):
+        config = controller.config.copy()
+        # Creat dict of required hardware instances.
+        hardware = {x: y for x, y in self.hardware.items() if x in config}
+        # Update controller config with actual hardware instances.
+        config.update(hardware)
+        # Instantiate controller and add to list.
+        self.controllers.append(controller.driver(**config))
```
jamienoss, line 86: @amitschang so like this?

amitschang: If I am reading this correctly, this seems to tie the config name to the hardware which would mean that a user can inadvertently override a config with a hardware, in turn it would mean the implementer might have to be careful about config naming. For example if a controller had a `temperature` setting (in degrees for example) and someone set some hardware name to `temperature`, this would override that unintentionally. That is not great example, but point is a-priori we couldn't know what list of names to exclude from config.

Perhaps we can do it with a set of "reserved" configs, like "hardwares": ["list", "of", "names", "or"] (or even just None to pass in all box hardware), and and "history": None (to get default), or the like. We'd have to be a little careful with future-proofing in case there was some other global/standard stuff needing passing.

jamienoss: >If I am reading this correctly, this seems to tie the config name to the hardware

There's a chance I too am not entirely following, but this sounds like the very point that I tried to raise yesterday.

The design so far is to hardcode the device/hardware names as attributes of both ``Controller`` and ``Controller.Config`` - which I really like. With our constructor pattern from a descriptor (json/str), these will always be strings and it's up to the classes validation of its own config model/``__init__`` to conduct this check on correctness, no?

The crux of the change above (which I'm not saying is correct - just alliterating as code for you so we can discuss) is that the controller should transparently/directly get what it needs (for encapsulation and being self-contained etc etc - aka OO stuff), but when instantiated from the evolver/manager the device/hardware list is not specified in the ``"controller"`` section of the config, so we either have to duplicate that or get the device list from the list of devices passed to the manager to manage. To use the manager pattern, the point is to assume/require that it holds/owns/manages all devices. What we don't want to do is blindly pass all controllers all hardware that the manager owns, right, so we need to filter.

jamienoss: So yes the config names for the hardware/devices are tethered to the controller attribute names - by design.
Yes, these aren't necessarily namespaced by controller and are shared in the evolver config - which, yes, they might not want to be, or be able to be as they won't be unique and may need to be.

Eitherway, this design issue is one that we need to address, as it still exists even if we just pass the evolver/manager instance to the controllers (which I really don't think we should do - because OO).

amitschang: Could we achieve the desired behavior by enabling a `HardwareDriverDescriptor` either get a name or the full class/config spec and then using that as a hook-in when building the whole evolver from config. So imagning (assuming some default for illustration):

```py
class MyController(Controller):
    class Config(BaseConfig):
        od: HardwareDriverDescriptor = HardwareDriverDescriptor(name='od_A')
```

and then the evolver part you have above would along lines of:

```py
def is_hardware_config_pluggable(config, name):
    if name not in config:
        return False    
    try: 
        return config.get(name).needs_driver()
    except:
        return False

plugged_hardware = {name: driver for name,driver in self.hardware.items() if is_hardware_config_pluggable(config, name)}
config.update(plugged_hardware)
```

So without changing the code I can switch out `od_A` for `od_B` in on-disk config. Similar hookin for History, etc. No need for any reserved or magical config names.

jamienoss: See #85

jamienoss:
No team buy-in so closing.

PR #68, opened 2024-05-21
Fix serial is_open check and response exception

amitschang:
A couple bug fixes. Turns out pyserial implements `is_open` as a property. The exception for bad response hid the response data since the walrus was assigned a boolean.

### DIFF
--- a/evolver/connection/interface.py
+++ b/evolver/connection/interface.py
@@ -48,7 +48,7 @@ class Connection(BaseInterface):
 
     def is_open(self):
         if hasattr(self.conn, "is_open"):
-            return self.conn.is_open()
+            return self.conn.is_open
 
         return self.conn is not None
 
--- a/evolver/connection/interface.py
+++ b/evolver/connection/interface.py
@@ -48,8 +48,7 @@ class Connection(BaseInterface):
 
     def is_open(self):
         if hasattr(self.conn, "is_open"):
-            return self.conn.is_open
-
+            return self.conn.is_open() if callable(self.conn.is_open) else self.conn.is_open
         return self.conn is not None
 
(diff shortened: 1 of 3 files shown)


evolver/serial.py
```diff
@@ -34,7 +34,7 @@ def _close(self):
     @classmethod
     def _decode_serial_data(cls, response, suffix=CMD.RESP_SUFFIX.value):
         parts = response.split(b',')
-        if resp_suffix := parts.pop() != suffix:
+        if (resp_suffix := parts.pop()) != suffix:
```
jamienoss, line 37: 👍

evolver/connection/interface.py
```diff
@@ -48,7 +48,7 @@ def communicate(self, *args, **kwargs):
 
     def is_open(self):
         if hasattr(self.conn, "is_open"):
-            return self.conn.is_open()
+            return self.conn.is_open
```
jamienoss, line 51:
```suggestion
            return self.conn.is_open() if callable(self.conn.is_open) else self.conn.is_open
```
I see your point that this could be a property, but it might not be since this is in the base interface, so we should check if callable etc.

jamienoss (changes requested):
Good catch! Thanks

evolver/connection/interface.py
amitschang: I couldn't apply directly for some reason, but made this change, thanks!

jamienoss: My bad, I copied and pasted it from a previous post attempt as I also tried to include the ``if`` to suggest using a ``getattr`` rather than ``hasattr`` but GitHub wouldn't let me make a suggestion stradling the deleted line. Perhaps, I added it back to the wrong line. 🤷

PR #71, opened 2024-05-23
Make all config fields required for app update_config path func

### LINKED ISSUE #70: Bug in app update config
See #69

jamienoss:
Resolves #70, #69

### DIFF
--- a/evolver/app/main.py
+++ b/evolver/app/main.py
@@ -1,11 +1,19 @@
 import asyncio
+
 from fastapi import FastAPI
+
+from evolver.base import require_all_fields
 from evolver.device import Evolver, EvolverConfig
 from .. import __project__, __version__
 
 
 app = FastAPI()
 evolver = Evolver()
+
+
+@require_all_fields
+class EvolverConfigWithoutDefaults(EvolverConfig):
+    ...
 
 
 @app.get("/")
@@ -26,7 +34,7 @@ async def get_state():
 
 
 @app.post("/")
-async def update_evolver(config: EvolverConfig):
+async def update_evolver(config: EvolverConfigWithoutDefaults):
     evolver.update_config(config)
 
 
--- a/evolver/app/tests/test_app.py
+++ b/evolver/app/tests/test_app.py
@@ -43,11 +43,7 @@ class TestApp:
             assert content["msg"] == "Field required"
 
         new_data = EvolverConfig().copy(update=data)
-        # Note: pydantic.BaseModel.model_dump() doesn't serialize the driver class and canonical json has issues
-        # serializing some classes, e.g., ABCMeta. However, pydantic.BaseModel.model_dump_json() knows how to convert
-        # pydantic.ImportString. However, some setting must be adrift as this results in a 422 "Input should be a valid
-        # dictionary or object to extract fields from". To resolve this, we convert it back to a dict.
-        response = app_client.post('/', json=json.loads(new_data.model_dump_json()))
+        response = app_client.post('/', data=new_data.model_dump_json())
         assert response.status_code == 200
         newconfig = app_client.get('/').json()['config']
         assert newconfig['hardware']['test']['driver'] == 'evolver.hardware.demo.NoOpSensorDriver'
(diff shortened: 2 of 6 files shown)


evolver/app/main.py
```diff
@@ -8,6 +11,11 @@
 evolver = Evolver()
 
 
+@require_all_fields
+class EvolverConfigWithoutDefaults(EvolverConfig):
```
jamienoss, line 15: I'm happy to name this something else.

evolver/device.py
```diff
@@ -32,8 +32,8 @@ class EvolverConfig(pydantic.BaseModel):
     vials: list = list(range(16))
     hardware: dict[str, HardwareDriverDescriptor] = {}
     controllers: list[ControllerDescriptor] = []
-    serial: ControllerDescriptor = None
-    history: ControllerDescriptor = None
+    serial: ControllerDescriptor = ControllerDescriptor(driver=DEFAULT_SERIAL, config=DEFAULT_SERIAL.Config().model_dump())
+    history: ControllerDescriptor = ControllerDescriptor(driver=DEFAULT_HISTORY, config=DEFAULT_HISTORY.Config().model_dump())
```
jamienoss, line 36: Smuggle to get tests working.

evolver/history.py
```diff
@@ -18,7 +18,9 @@ def get(self, query):
 
 
 class HistoryServer(History):
-    def __init__(self, config = None):
+    def __init__(self, evolver=None, config=None):
+        self.evolver = evolver
+        self.config = config
```
jamienoss, line 23: Smuggle to get tests working.

evolver/app/tests/test_app.py
```diff
@@ -25,9 +29,25 @@ def test_evolver_app_default_config_dump_endpoint(self, app_client):
...
+        contents = json.loads(response.content)
+        for content in contents["detail"]:
+            assert content["msg"] == "Field required"
+
+        new_data = EvolverConfig().copy(update=data)
+        # Note: pydantic.BaseModel.model_dump() doesn't serialize the driver class and canonical json has issues
+        # serializing some classes, e.g., ABCMeta. However, pydantic.BaseModel.model_dump_json() knows how to convert
+        # pydantic.ImportString. However, some setting must be adrift as this results in a 422 "Input should be a valid
+        # dictionary or object to extract fields from". To resolve this, we convert it back to a dict.
+        response = app_client.post('/', json=json.loads(new_data.model_dump_json()))
```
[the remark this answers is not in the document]
jamienoss: ``app_client.post('/', json=json.dumps(json.loads(new_data.model_dump_json())))`` also gives a 422 so looks like a ``post`` issue.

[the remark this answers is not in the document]
jamienoss: Yep, https://github.com/tiangolo/fastapi/issues/192.

The docs are very misleading, especially since [the docs](https://fastapi.tiangolo.com/reference/testclient/?h=testclient#fastapi.testclient.TestClient.post) state that ``json`` can be of any type. Now, I'm getting Deja Vu and remember hitting this issue in the past.

evolver/base.py
```diff
@@ -2,6 +2,33 @@
...
+
+            class MyModel(pydantic.BaseModel):
+                a: int = 3
+
+            @require_all_fields
+            class MyModelWithoutDefaults(MyModel):
+                ...
+            ```
+    """
+    for field in cls.model_fields:
```
jamienoss, line 24: I looked to see if there was a more pydantic/fastapi way of doing this, but couldn't find anything canonical so came up with this. If there is a more correct way of doing this within the framework, I'd rather use that than this.

Another option is to create a child class of ``FieldInfo`` with ``is_required`` overridden to always return ``True``, however, I thought it simple to mutate the attrs on the single model class than create new class instances for each ``Field``.

evolver/app/tests/test_app.py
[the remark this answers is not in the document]
amitschang: I think this matches the call for requests, where the json field is processed into json (from a python object), so you could probably achieve the same thing as here (without multiple serializations) with:

```py
 response = app_client.post('/', data=new_data.model_dump_json())
```

amitschang (commented):
couple comments/questions

amitschang (approved):

Issue #22, opened 2024-04-04
create implementations for default sensors/effectors


amitschang:
Including:
* [x] temp
* [x] pumps
* [x] od led
* [x] od photodiode
* [x] stirrer

reference: https://github.com/FYNCH-BIO/evolver/

PR #72, opened 2024-05-23
Add initial OD sensor implementation

amitschang:
Part of https://github.com/ssec-jhu/evolver-ng/issues/22

Adds an OD sensor implementation which corresponds to that on the default evolver box. For now the calibration is left out but this shows clearly where it would fit: in providing a value for "density" field of "Output".

Adding also some shared infrastructure for testing these serial-based devices. Probably more of the test infra could be shared for hardware modules, but leaving it specific to OD here.

### DIFF
--- a/evolver/hardware/standard/od_sensor.py
+++ b/evolver/hardware/standard/od_sensor.py
@@ -1,5 +1,6 @@
 from pydantic import Field
 
+from evolver.connection.interface import Connection
 from evolver.hardware.interface import SensorDriver, VialBaseModel, VialConfigBaseModel
 from evolver.serial import SerialData
 
@@ -14,16 +15,21 @@ class ODSensor(SensorDriver):
     class Config(VialConfigBaseModel):
         addr: str = Field(description="Address of od sensor on serial bus (e.g. od_90)")
         integrations: int = Field(500, description="on read, request average of this number of ADC reads")
+        serial_conn: Connection | None = Field(None, description="serial connection, default is that on evolver")
 
     class Output(VialBaseModel):
         raw: int
         density: float = None
 
+    @property
+    def serial(self):
+        return self.serial_conn or self.evolver.serial
+
     def read(self):
         self.outputs = {}
         cmd_data = str(self.integrations).encode()
         cmd = SerialData(addr=self.addr, data=[cmd_data], kind="r")
-        with self.evolver.serial as comm:
+        with self.serial as comm:
             response = comm.communicate(cmd)
         for vial, raw in enumerate(response.data):
             if vial in self.vials:
--- a/evolver/hardware/standard/od_sensor.py
+++ b/evolver/hardware/standard/od_sensor.py
@@ -1,0 +1,27 @@
+from pydantic import Field
+from evolver.hardware.interface import SensorDriver, VialBaseModel, VialConfigBaseModel
+from evolver.serial import SerialData
+
+
+class ODSensor(SensorDriver):
+    """Optical density sensor driver
+
+    This driver represents a family of sensors for turbidity which are read by
+    integrating a specified number of ADC readings per vial.
+    """
+    class Config(VialConfigBaseModel):
+        addr: str = Field(description="Address of od sensor on serial bus (e.g. od_90)")
+        integrations: int = Field(500, description="on read, request average of this number of ADC reads")
+
+    class Output(VialBaseModel):
+        raw: int
+        od: float = None
+
+    def read(self):
+        self.outputs = {}
+        cmd_data = str(self.config.integrations).encode()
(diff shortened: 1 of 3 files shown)


evolver/hardware/standard/od_sensor.py
```diff
@@ -0,0 +1,35 @@
...
+
+    class Output(VialBaseModel):
+        raw: int
+        density: float = None
+
+    @property
+    def serial(self):
+        return self.serial_conn or self.evolver.serial
+
+    def read(self):
```
jamienoss, line 28: BTW, when I was updating #51 last week I realized that there's an awkwardness in the API, in that ``read`` and ``get`` are separate, but more importantly that ``read`` doesn't return. I think it would be better if it did. Not necessarily a suggestion to smuggle into this PR but just an FYI.

evolver/hardware/standard/od_sensor.py
```diff
@@ -0,0 +1,35 @@
...
+        self.outputs = {}
+        cmd = SerialData(addr=self.addr, data=[str(self.integrations).encode()], kind="r")
+        with self.serial as comm:
+            response = comm.communicate(cmd)
+        for vial, raw in enumerate(response.data):
+            if vial in self.vials:
+                self.outputs[vial] = self.Output(vial=vial, raw=raw)
```
jamienoss, line 35: If ``self.vials`` is updated once live, and previous values are taken and thus ``self.outputs`` is already populated, who's responsibility is it to set/reset ``self.outputs``? Here we only set the ones we think we want, but ``get`` returns ``self.outputs`` which could include stale data. It may be better to address that here.

evolver/hardware/standard/od_sensor.py
```diff
@@ -0,0 +1,35 @@
...
+    @property
+    def serial(self):
+        return self.serial_conn or self.evolver.serial
+
+    def read(self):
+        self.outputs = {}
+        cmd = SerialData(addr=self.addr, data=[str(self.integrations).encode()], kind="r")
+        with self.serial as comm:
+            response = comm.communicate(cmd)
+        for vial, raw in enumerate(response.data):
```
jamienoss, line 33: Remind me, is it guaranteed that this is ordered by vial such that we can just enumerate here?

evolver/serial.py
```diff
@@ -69,17 +69,23 @@ def communicate(self, cmd: SerialData):
...
         self._data = None
+        self._data_raw = None
 
     def write(self, data):
+        self._data_raw = data
         self._data = EvolverSerialUART._decode_serial_data(data, suffix=EvolverSerialUART.CMD.SEND_SUFFIX.value)
 
     def readline(self):
+        if mapped_response := self.raw_response_map.get(self._data_raw):
+            return mapped_response
```
jamienoss, line 88: Why shortcircuit the emulator response handling?

evolver/hardware/standard/od_sensor.py
jamienoss: It would be good to add a test for this.

amitschang: Another read would have output only include a new set of vials, I can add a test for this

evolver/hardware/standard/od_sensor.py
amitschang: yes, it will come back in order - in fact the position in the response is the only indicator of vial number

evolver/serial.py
amitschang: not sure I get the question - the added response mapping for testing depends on this. It enables the tests added here that use `serial_mock` fixture.

evolver/hardware/standard/od_sensor.py
amitschang: ah, indeed `get` data are sort of by definition stale (its like get access to previous read). Would have to think about what it would mean to update the vial list in place and whether that should affect the gets before next read.

jamienoss: >Would have to think about what it would mean to update the vial list in place and whether that should affect the gets before next read.

I agree, but to me, this just highlights a root issue with separating `read`` and ``get`` like this. It doesn't seem intuitive and also requires additional semantics and functionality to make it correct. As also noted above in this PR https://github.com/ssec-jhu/evolver-ng/pull/72#discussion_r1633264022.

amitschang: ok, that is another thing, we can `get` to that later 🤣 ,  I'll leave the behavior as-is for this change.

jamienoss: ba dum tshh

evolver/serial.py
jamienoss: Ah, I see now, took another look and see what's going on.

jamienoss (approved):

PR #76, opened 2024-05-28
Post review changes to #74

### LINKED ISSUE #67: Global evolver instance at the app layer
Currently, we have:
```python
# evolver.app.main

app = FastAPI()
evolver = Evolver()
```

This won't work when running the app from uvicorn using multiple workers, they'll each get their own copy that won't be the same instance. It's possible that this may not matter, however, I expect that multiple serial connections will not work. Oh, and the main loop def won't work as desired.

For the app on the box, we probably don't want to run with multiple workers anyhow.

Running the app elsewhere though, should still be ok as this would have to go indirectly using the web app connection rather than the serial connection.

From my understanding either way, the more correct FastAPI way to do this, is to stash it in ``app``, though I'm not yet sure of what happens there.
```python
# evolver.app.main

app = FastAPI()
app.evolver = Evolver()
```

jamienoss:
Improvements to #74
Resolves #67 

 - Adds ``save`` and ``load`` methods to ``evolver.base.BaseConfig``
 - Separate settings for application settings.
 - Reading the docs, it's now better to use a lifespan event rather than individual startup and shutdown events.
 - Instantiate app ``evolver`` instance on startup, not at module level. (This required #67 to be resolved - see https://github.com/ssec-jhu/evolver-ng/issues/67#issuecomment-2135687714)

See https://github.com/ssec-jhu/evolver-ng/pull/74#issuecomment-2135456108
>I think the files should rep descriptors not just the config such that the file also has the classinfo so that when looking at a config file you know what model it's for.

Also, we should note that ``pydantic`` has similar functionality for reading from files [YamlConfigSettingsSource](https://docs.pydantic.dev/latest/api/pydantic_settings/#pydantic_settings.YamlConfigSettingsSource) but isn't quite what we want.

### DIFF
--- a/evolver/app/tests/conftest.py
+++ b/evolver/app/tests/conftest.py
@@ -1,10 +1,17 @@
-import os
 import pytest
 from fastapi.testclient import TestClient
-from ..main import app  # Leave as relative for use in template: ssec-jhu/base-template.
+
+from evolver.app.main import app
+from evolver.device import EvolverConfig
+from evolver.settings import app_settings
 
 
-@pytest.fixture(scope="class")
-def app_client(tmp_path_factory):
-    os.chdir(tmp_path_factory.mktemp('test_dir'))
-    return TestClient(app)
+@pytest.fixture(scope="function")
+def app_client(tmp_path, monkeypatch):
+    monkeypatch.setattr(app_settings, "CONFIG_FILE", tmp_path / app_settings.CONFIG_FILE)
+
+    # Create and save a default config file to be read upon app startup.
+    EvolverConfig().save(app_settings.CONFIG_FILE)
+
+    with TestClient(app) as client:
+        yield client
--- a/evolver/app/tests/conftest.py
+++ b/evolver/app/tests/conftest.py
@@ -6,7 +6,7 @@ from evolver.device import EvolverConfig
 from evolver.settings import app_settings
 
 
-@pytest.fixture(scope="function")
+@pytest.fixture
 def app_client(tmp_path, monkeypatch):
     monkeypatch.setattr(app_settings, "CONFIG_FILE", tmp_path / app_settings.CONFIG_FILE)
 
--- a/evolver/app/tests/test_app.py
+++ b/evolver/app/tests/test_app.py
@@ -1,11 +1,13 @@
-import os
 import json
 import yaml
+
 from fastapi.openapi.utils import get_openapi
 from fastapi.testclient import TestClient
+
+from evolver import __version__
 from evolver.settings import app_settings
 from evolver.device import EvolverConfig, HardwareDriverDescriptor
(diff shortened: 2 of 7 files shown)


jamienoss:
@amitschang converted this back to a draft for a mo, whilst I sort out the tests and config file.

evolver/app/tests/conftest.py
```diff
@@ -1,10 +1,17 @@
...
 import pytest
 from fastapi.testclient import TestClient
-from ..main import app  # Leave as relative for use in template: ssec-jhu/base-template.
 
+from evolver.app.main import app
+from evolver.device import EvolverConfig
+from evolver.settings import app_settings
 
-@pytest.fixture(scope="class")
-def app_client(tmp_path_factory):
```
jamienoss, line 8: ``tmp_path_factory`` is session scoped whilst ``tmpdir`` and ``tmp_path`` are test invocation scoped.

evolver/settings.py
```diff
@@ -12,6 +12,7 @@ class Settings(BaseSettings):
 
 class AppSettings(BaseSettings):
     CONFIG_FILE: Path = Path('evolver.yml')  # in current directory
+    LOAD_FROM_CONFIG_ON_STARTUP: bool = True
```
jamienoss, line 15: I added this as I imagine this will want to be configurable and explicit such that it doesn't silently not load from config when the user might expect it to.

evolver/app/tests/test_app.py
```diff
@@ -62,12 +64,11 @@ def test_evolver_app_control_loop_setup(self, app_client):
             assert response.status_code == 200
 
 
-def test_app_load_file(tmp_path):
-    os.chdir(tmp_path)
```
jamienoss, line 67: Monkey patching ``app_settings.CONFIG_FILE`` would be a better way to do this.

amitschang (approved):

evolver/util.py
```diff
@@ -14,16 +11,3 @@ def find_package_location(package=__project__):
 
 def find_repo_location(package=__project__):
     return Path(find_package_location(package) / os.pardir)
-
-
-def load_config_for_evolver(evolver: Evolver, configfile: Path):
-    """Loads the specified config file and applies to evolver."""
-    if configfile.exists():
```
jamienoss, line 21: @amitschang My removing this so that this functionality is explicit, compounded with the mokeypatch that I added instead of your ``chdir`` caused the update app test to fail. This also then made me realize, as I had already brought to your attention, the scope of the testclient fixture was incorrect, changing that here such that the config loading and updating test was actually testing what we expected and not a previously saved config from another test, then required the evolver to be instantiated at runtime (app startup) and not import time. That then lead to the lifespan addition since order between instantiation and config loading could be better addressed, which then lead to adding it to the app, so that the instantiation from within a func was more elegant.

amitschang (approved):

PR #83, opened 2024-06-06
Don't duplicate log handlers

### LINKED ISSUE #82: interface logging handler duplication
The base interface does some logging setup, but in case there are two things of the same type or name, we can get duplicates of the handler, for example, if `MySensor` is descendent of `BaseInterface`

```py
x = MySensor()
y = MySensor()
print(x.logger.handlers)
print(y.logger.handlers)
```

what we see is:
```py
[<StreamHandler stderr (INFO)>, <StreamHandler stderr (INFO)>]
[<StreamHandler stderr (INFO)>, <StreamHandler stderr (INFO)>]
```

because a reconfigure creates new instances in a running process (so that the logger sticks around) , this will likely lead to a lot of increasingly duplicate logs.

It's nice to have the `logger` in there, but I suspect that most cases we'll want to set up the handlers/formatters in root a certain way and have the loggers derive from there, and otherwise they should probably be set explicitly by the user - so the `BaseInterface` could prob refrain from setting the handler at all.

jamienoss:
Resolves #82

This also moves the log level and format to configurable settings.

### DIFF
--- a/evolver/app/main.py
+++ b/evolver/app/main.py
@@ -3,10 +3,14 @@ from contextlib import asynccontextmanager
 
 from fastapi import FastAPI
 
+import evolver.util
 from evolver import __project__, __version__
 from evolver.base import require_all_fields
 from evolver.device import Evolver
 from evolver.settings import app_settings
+
+# Setup logging.
+evolver.util.setup_logging()
 
 
 @asynccontextmanager
--- a/evolver/base.py
+++ b/evolver/base.py
@@ -7,7 +7,6 @@ import pydantic
 import pydantic_core
 import yaml
 
-from evolver.settings import settings
 import evolver.util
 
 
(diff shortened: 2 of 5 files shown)


jamienoss:
@amitschang we can always readdress the logging design later, i.e., single global vs individual per device. Either way, this PR will at least patch the bug and stop duplicate log messages.

amitschang (commented):
This gets at one source of duplication, but log messages still may pass through multiple handlers in an unexpected way. For example, try this:

```py
import logging
logging.basicConfig()
a = ConcreteInterface()
a.logger.error('error')
```

and you may see something quite like this:
```
2024-06-07 10:50:54,678 - ConcreteInterface - ERROR - error
ERROR:ConcreteInterface:error
```

Typically a library just gets loggers and does logging, the application portion being the one that sets up how to handle it. Here this is library-side, the place where we start application (e.g. in `app`) should do the handler setup. And in most cases I would suspect that be setup at the root level.

I think removing the handler from here is more correct

jamienoss:
>Typically a library just gets loggers and does logging,

Yep, solid point.

amitschang (approved):

PR #47, opened 2024-04-25
Refactor hardware interface

jamienoss:
This refactors the base hardware interface.

Due to the coupling between the device interface and the calibrator, this PR also implements the calibrator interface.

Resolves https://github.com/ssec-jhu/evolver-ng/issues/44

Issue #45, opened 2024-04-24
Bulk out interface for Calibrator.run_calibration_procedure


jamienoss:
Related to #44 & #36

E.g.:

 * dumping calibration files to storage
 * stuff needed for web app
 * could this run an instance of ``Controller``?
 * 🤔

amitschang:
We have this now, started in #175 and evolved over several PRs and issues

PR #86, opened 2024-06-13
Add calibration interface

### LINKED ISSUE #44: Add calibrator interface
Relates to #36 and #28

jamienoss:
Resolves #44 (partially, not inc #45)

Moved out of #47

The final design pattern is for the convert and fitting functionality to belong to the ``Transformer`` interface, and then the ``Calibrator`` interface takes these as config params for both ``input_transformer`` and ``output_transformer``, along with this interface housing the runtime calibration procedure.


evolver/calibration/interface.py
```diff
@@ -0,0 +1,64 @@
...
+
+    class Config(BaseConfig):
+        dir: Path = Field(
+            settings.ROOT_CALIBRATOR_FILE_STORAGE_PATH, description="Directory for saving new calibration files."
+        )
+
+        def save(self, file_path: Path = None, encoding: str | None = None):
+            if file_path is None:
+                file_path = Path(
+                    f"{self.name}_{datetime.datetime.now().strftime(settings.DATETIME_PATH_FORMAT)}"
```
jamienoss, line 26: Should we always append a timestamp, not just for when no filename is given?

evolver/calibration/demo.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def convert(
+        self, data: NoOpSensorDriver.Output | dict[Any, NoOpSensorDriver.Output]
+    ) -> NoOpSensorDriver.Output | dict[int, NoOpSensorDriver.Output]:
+        """Inplace calibration conversion."""
+
+        def _convert(x):
+            return x * self.conversion_factor + self.conversion_constant
+
+        if isinstance(data, dict):
```
jamienoss, line 39: It looks to me that there's scope to either improve the interface coupling between ``Calibrator`` and ``HardwareDriver`` in respect to whether ``HardwareDriver.get()`` always returns a ``dict`` or we can common up some of this functionality so that the user/dev doesn't have to add this boiler plate all the time.

jamienoss:
Update: We've discussed offline to adopt some of the design aspects I proposed a month or so ago, so this PR is still WIP.

The design points are:
- Remove ``is_calibrated`` semantics as this really isn't achievable - to really know whether calibration values are still accurate or not is not possible without running the calibration procedure to find out. Instead, we'll add timestamps to the calibration values and a mechanism to:
  - warn the user that these are old and could be inaccurate after some configurable time has elapsed
  - warn the user upon first use that they are using the shipped default values.
- Decompose "conversion" and "calibration" semantics, i.e., "conversion" is that between raw data and device units, e.g., voltage and temp in degrees C for a thermometer device. This yields a few advantages such as:
  - allows ``Output`` models to always have populated both ``raw`` and ``value`` fields, currently, this is not possible since ``value`` field(s) are those "calibrated" which may not have taken place yet. The ``Output`` classes also belong to the hardware interface and not the calibrator interface and thus shouldn't be entangled.
  - allows for separate semantics & calibration implementations between absolute and relative calibrations. E.g., the linear relation between a thermometer's output voltage/resistance and temperature response is an absolute calibration that is unlikely to ever change, however, relative calibration between thermometers mounted on separate vials is vial dependent (thus as likely to change as smart sleeves are swapped - if at all) and really only requires just a calibration offset, not an offset and gradient as is needed for the V-temp relation (assuming it's linear).

evolver/settings.py
```diff
@@ -13,6 +15,13 @@ class Settings(BaseSettings):
     DEFAULT_NUMBER_OF_VIALS_PER_BOX: int = 16
     LOG_LEVEL: str = "INFO"
     LOG_FORMAT: str = "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
+    ROOT_CALIBRATOR_FILE_STORAGE_PATH: Path = Path("calibration_files")
+    DEFAULT_CALIBRATION_EXPIRE: timedelta = Field(
+        timedelta(weeks=4 * 6),  # Approx. 6 months.
```
jamienoss, line 20: Should we make this ``None``?

evolver/tests/test_device.py
```diff
@@ -36,6 +37,11 @@ def demo_evolver(conf_with_driver):
 
 
 class TestEvolver:
+    def test_demo_evolver(self, demo_evolver, conf_with_driver):
```
jamienoss, line 40: FYI This test was added to debug #104 and I thought it best to leave it in.

evolver/calibration/demo.py
```diff
@@ -0,0 +1,17 @@
...
+
+    convert_from = convert_to
+
+
+class NoOpCalibrator(Calibrator):
+    class Config(Calibrator.Config):
+        input_transformer: ConfigDescriptor | Transformer | None = NoOpTransformer()
+        output_transformer: ConfigDescriptor | Transformer | None = NoOpTransformer()
+
+    def run_calibration_procedure(self, *args, **kwargs): ...
```
jamienoss, line 17: With the addition of the ``LinearCalibrator``, etc, these demo classes seem quite redundant at this point - especially since they don't seem to demo much.

It seems that their only merit is their use in tests to de-tether breakages in ``polyfit`` causing cascading breakages in other unit tests, thus making them less unit-y.

evolver/base.py
```diff
@@ -94,6 +98,22 @@ def shallow_model_dump(self):
...
 
+CreatedTimestampField = partial(
+    pydantic.Field, description="The creation timestamp", default_factory=datetime.datetime.now
+)
+ExpireField = partial(
+    pydantic.Field,
+    default=None,
+    description="The amount of time after which the associated object is considered stale. "
+    "`None` := forever (the default).",
+)
```
jamienoss, line 109: I feel like there should be a more accepted way of doing this, i.e., using ``Annotated``. However, I tried this but couldn't get ``pydantic.Field`` param ``default_factory`` to work.

evolver/settings.py
jamienoss: Also, arguably this should use ``evolver.base.ExpireField``, however, that would result in circular imports as soon as ``evolver.base`` imported settings (which is likely).

jamienoss: Could move the dec to here instead.

evolver/calibration/standard/polyfit.py
```diff
@@ -0,0 +1,59 @@
...
+class PolyFitTransformer(Transformer):
+    class Config(Transformer.Config):
+        degree: int = Field(ge=0, description="Polynomial degree.")
+        coefficients: list[float] = Field(min_length=1, description="Polynomial coefficients.")
+
+        @model_validator(mode="after")
+        def check_coefficients_length(self) -> Self:
+            if len(self.coefficients) - 1 != self.degree:
+                raise ValueError(f"Degree={self.degree} but {len(self.coefficients)} coefficients given.")
+            return self
```
amitschang, line 19: it doesn't really appear degree from config is used at all, or does this somehow get fed from config params into the `fit` function? I might just be missing something, or degree from config is not consumed outside this check?

evolver/calibration/standard/polyfit.py
```diff
@@ -0,0 +1,59 @@
...
+    def convert_from(self, y):
+        return super().convert_from(y)[0]
+
+
+class PolyFitCalibrator(Calibrator):
+    def run_calibration_procedure(self, data: dict):
+        # TODO: This is just a placeholder.
+        calibration_data = {}
+        for transformer, (x, y) in data.items():
+            calibration_data[transformer] = getattr(self, transformer).refit(x, y)
```
amitschang, line 51: Does `refit` here need the degree?

evolver/device.py
```diff
@@ -42,7 +42,13 @@ def effectors(self):
 
     @property
     def calibration_status(self):
-        return {name: device.calibrator.status for name, device in self.hardware.items()}
+        """Return the calibration Status for each device's calibrator. This is explicit and returns all available
```
amitschang, line 45: perhaps `Status for each device's calibrator.` -> `Status for each device's calibrator, or None if hardware has no calibrator.`

evolver/tests/test_device.py
```diff
@@ -126,3 +132,10 @@ def test_non_descriptor_config_field(self):
         assert isinstance(obj.hardware["a"], NoOpSensorDriver)
         assert isinstance(obj.serial, DEFAULT_SERIAL)
         assert isinstance(obj.history, DEFAULT_HISTORY)
+
+    def test_calibration_status(self, demo_evolver):
+        status = demo_evolver.calibration_status
+        assert status.keys() and (status.keys() == demo_evolver.hardware.keys())
+        assert status["testeffector"] is None
+        assert isinstance(status["testsensor"]["input_transformer"], Status)
+        assert isinstance(status["testsensor"]["output_transformer"], Status)
```
amitschang, line 141: Looking at this, wondering if it might make sense to have the calibrator have an overall status, so maybe it returns an object (effectively) like:

```py
class CalibratorStatus:
    input_transformer_status: Status
    output_transformer_status: Status
    ok: bool = input.ok and output.ok
```

thoughts?

amitschang (commented):
Some inline comments. 

Also have one overall though/question: From my read it looks like a primary target for doing the procedure would be establishing and saving the parameters of the fit (transformer). It does look, however, like it is also amenable to having the calibrator store the sets of raw voltage and target value during its procedure (for temp for instance) and then in it's constructor (and/or after procedure) call the `refit` of a transformer from those. I think the latter is probably more flexible method in general (can switch poly fit degree after the fact, for example) - I guess question is, did I get that right? Would there be anything making the second method non-trivial?

evolver/calibration/standard/polyfit.py
jamienoss: Thanks, good catch. I missed the ``refit`` overrides needed when I converted ``fit`` to a ``classmethod``.

evolver/calibration/standard/polyfit.py
jamienoss: See https://github.com/ssec-jhu/evolver-ng/pull/86/files#r1711861266.
Handled by ``refit`` overrides such that this correctly belongs to the transformer rather than being the responsibility of the calibrator.

evolver/app/tests/test_app.py
```diff
@@ -91,6 +96,87 @@ def test_schema_endpoint_exception(self, app_client, classinfo):
...
+        response = app_client.get("/calibration_status/", params=dict(name=None))
+        assert response.status_code == 200
+        contents = json.loads(response.content)
+
+        for device in hardware:
+            assert contents[device]
+            for transformer in ("input_transformer", "output_transformer"):
+                status = Status.model_validate(contents[device][transformer])
+                assert status.ok
+                assert status.created - datetime.now() < timedelta(seconds=5)
```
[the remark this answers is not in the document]
jamienoss: This is actually a much better test, thanks. There was a bug in ``NoOpCalibrator`` where the defaults were being created at import time (and thus shared instances) and not runtime with the use of ``Field(default_factory=NoOpTransformer)``.

evolver/tests/test_device.py
jamienoss: I don't think this can be utilized as a return type for ``app.main.calibration_status``, however, as it can be customized by derived calibrator classes. Since, for example, ``input_transformer`` belongs to the ``Config`` it can be deleted and other fields added, in which case ``Status`` will also need to be updated to account for these new fields. I had originally contemplated this, however the above dissuaded me. This is also why ``Status`` needs to belong to ``Calibration`` rather than at the module level.

amitschang (approved):
Looks good, thanks for this and the updates!

PR #99, opened 2024-07-10
Only shallow model dump from ConfigDescriptor.create to keep nested descriptors as instances and not dicts

jamienoss:
Discovered during #86 where hardware calibrator descriptors wouldn't get instantiated as they were being converted back to ordinary dictionaries during the nested creation process from ``Evolver.create()``.

### DIFF
--- a/evolver/tests/test_base.py
+++ b/evolver/tests/test_base.py
@@ -4,6 +4,7 @@ import pydantic
 import pytest
 
 import evolver.base
+import evolver.util
 
 
 class ConcreteInterface(evolver.base.BaseInterface):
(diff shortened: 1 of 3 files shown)


evolver/tests/test_base.py
```diff
@@ -280,6 +286,29 @@ def test_from_file(self, tmp_path, mock_descriptor):
         assert obj.a == 11
         assert obj.b == 22
 
+    @pytest.mark.parametrize("shallow", (True, False))
+    def test_nested(self, monkeypatch, shallow):
+        if not shallow:
+            monkeypatch.setattr(evolver.base._BaseConfig, "shallow_model_dump", lambda x: x.model_dump())
+
```
amitschang, line 293: why is this necessary, since we don't really have a formal option of shallow-or-deep?

amitschang (approved):
Looks good, but wondering about something in test may or may not warrant a change

evolver/tests/test_base.py
jamienoss: Yeah, it's a bit odd I agree, I wanted to add a test to show when this fails to ensure that the test does actually test the success since it's a bit of a narrow gap to test.

``model_dump()`` is "deep" so I monkey patch it back to a deep conversion to ensure that it fails. If it doesn't fail, the test isn't adequate.

PR #100, opened 2024-07-11
pump and stir

amitschang:
Implement stirrer and pumps. 

In the case of stirrer this might not be correct in the case of calibration in that with calibration we might represent rate as RPM, but from the looks of their example experiment code the stirring is practically just used as some rate. Should be straightforward to update since this is a pretty simple effector.

For the pumps, the approach is to model a generic pump that is operates external to vials and indexes on pump number (48 of them), and supports a combination of standard and IPP pumps (these operate in different ways, but communicate to same underlying chip https://github.com/FYNCH-BIO/evolver-arduino/blob/master/SAMD21/RS485_FLUIDICS/RS485_FLUIDICS.ino), and then on top of this implement a per-vial influx/efflux setup that uses a generic device as a slave. A slave device rather than subclass is used due to different input types and mapping (2-to-1) for the vial driver.

### DIFF
--- a/evolver/hardware/standard/pump.py
+++ b/evolver/hardware/standard/pump.py
@@ -21,7 +21,7 @@ class GenericPump(EffectorDriver):
     non-IPP pumps.
     """
 
-    class Config(SerialDeviceConfigBase):
+    class Config(SerialDeviceConfigBase, EffectorDriver.Config):
         ipp_pumps: bool | list[int] = Field(False, description="False (no IPP), True (all IPP), or list of IPP ids")
 
     class Input(BaseConfig):
--- a/evolver/hardware/standard/pump.py
+++ b/evolver/hardware/standard/pump.py
@@ -86,7 +86,7 @@ class VialIEPump(EffectorDriver):
 
     class Input(VialBaseModel):
         flow_rate_influx: float = Field(description="influx flow rate in ml/s")
-        flow_rate_efflux: float | None = Field(None, description="efflux flow rate in ml/s. Defaults to same as influx")
+        flow_rate_efflux: float = Field(description="efflux flow rate in ml/s")
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -104,8 +104,6 @@ class VialIEPump(EffectorDriver):
     def commit(self):
         for p in self.proposal.values():
             if p.vial in self.vials:
-                if p.flow_rate_efflux is None:
-                    p.flow_rate_efflux = p.flow_rate_influx
                 self._generic_pump.set(GenericPump.Input(pump_id=self.influx_map[p.vial], flow_rate=p.flow_rate_influx))
                 self._generic_pump.set(GenericPump.Input(pump_id=self.efflux_map[p.vial], flow_rate=p.flow_rate_efflux))
         self._generic_pump.commit()
--- a/evolver/hardware/standard/pump.py
+++ b/evolver/hardware/standard/pump.py
@@ -1,0 +1,105 @@
+from copy import copy
+
+from pydantic import Field
+
+from evolver.base import BaseConfig
+from evolver.hardware.interface import EffectorDriver, VialBaseModel
+from evolver.hardware.standard.base import SerialDeviceConfigBase
+from evolver.serial import SerialData
+
+
+class GenericPump(EffectorDriver):
+    """Pump control for evolver fluidics directly by pump ID.
+
+    This driver controls either or both of standard and IPP pumps according to the
+    arduino fluidics module. Pumps are indexed by pump ID (either position in array or
+    IPP pump number) and not vial, each must be controlled separately.
+
+    Specify the index of any IPP pumps via the config parameter ipp_pumps, this should
+    be an array of those pump IDs that are IPP. Please note that IPP pumps reserve 3
+    slots (1 for each solenoid), so if pump 0 is IPP, 1 and 2 cannot be assigned as
+    non-IPP pumps.
+    """
+
+    class Config(SerialDeviceConfigBase):
+        ipp_pumps: list[int] = Field([], description="Indexes of IPP mode pumps, per solenoid")
+
+    class Input(BaseConfig):
+        pump_id: int
+        flow_rate: float
+
+    @property
+    def serial(self):
+        return self.serial_conn or self.evolver.serial
+
+    def set(self, input):
+        self.proposal[input.pump_id] = input
+
+    def commit(self):
+        inputs = copy(self.committed)
+        inputs.update(self.proposal.items())
+        cmd = [b"--"] * self.slots
+        for pump, input in inputs.items():
+            if pump in self.ipp_pumps:
+                # TODO: calibation transform here. The IPP pumps should return array of
+                # hz values per solenoid number - containing 3 of them.
+                hz_a = [input.flow_rate] * 3
+                for solenoid, hz in enumerate(hz_a):
+                    # solenoid is 1-indexed on hardware
+                    cmd[pump * 3 + solenoid] = f"{hz}|{pump}|{solenoid+1}"
+            elif any(pump - i < 3 for i in self.ipp_pumps):
+                raise ValueError(f"pump slot {pump} reserved for IPP pump, cannot address as standard")
+            else:
+                # TODO: calibration transform here. The time_to_pump is the target variable
+                # where pump_interval would presumably be what the pump_interval was set to
+                # during calibration and is fixed post-calibration.
+                time_to_pump, pump_interval = (input.flow_rate, int(input.flow_rate))
+                cmd[pump] = f"{time_to_pump}|{pump_interval}".encode()
+        with self.serial as comm:
+            comm.communicate(SerialData(addr=self.addr, data=cmd))
+        self.committed = inputs
+
+
+class VialIEPump(EffectorDriver):
+    """Vial-based Influx-Efflux Pump driver.
+
+    Each pump in the array is either an influx (to bring liquid to the vial) or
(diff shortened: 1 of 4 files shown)


evolver/hardware/standard/pump.py
```diff
@@ -0,0 +1,112 @@
...
+        efflux_map: dict[int, int] | None = Field(None, description="map of vial to efflux pump ID")
+
+    class Input(VialBaseModel):
+        flow_rate_influx: float = Field(description="influx flow rate in ml/s")
+        flow_rate_efflux: float | None = Field(None, description="efflux flow rate in ml/s. Defaults to same as influx")
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        p_slots = self.slots * 3  # influx, efflux, spare (or 3 solenoids for IPP)
+        self._generic_pump = GenericPump(
```
amitschang, line 94: In terms of calibration, wondering if this might be better as a separate hardware in evolver and this ones config has a reference to that. Not sure if the calibration would be done per-pump or as a 2-pump unit connected to a vial...

evolver/hardware/standard/pump.py
```diff
@@ -0,0 +1,112 @@
...
+            addr=self.addr,
+            slots=p_slots,
+            serial=self.serial_conn,
+            ipp_pumps=self.ipp_pumps,
+            evolver=kwargs.get("evolver"),
+        )
+        self.influx_map = self.influx_map or {i: i for i in range(0, p_slots)}
+        self.efflux_map = self.efflux_map or {i: i + self.slots for i in range(0, p_slots)}
```
jamienoss, line 102: FYI this kind of default inter-field initialization can happen in [model_post_init](https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_post_init).

We should discuss whether this is something we would prefer or even need, e.g., does the config need to be complete & correct without the need to instantiate its container?  Would this be needed for the UI, to render the efflux map from the config alone?

evolver/hardware/standard/pump.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def commit(self):
+        for p in self.proposal.values():
+            if p.vial in self.vials:
+                if p.flow_rate_efflux is None:
+                    p.flow_rate_efflux = p.flow_rate_influx
```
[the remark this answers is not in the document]
jamienoss: Do we need to add the alternate case so that it can be symmetric?

Should there be a flag for this behavior to allow the user to either raise or lower the fluid level?

evolver/hardware/standard/pump.py
```diff
@@ -0,0 +1,112 @@
...
+                hz_a = [input.flow_rate] * 3
+                for solenoid, hz in enumerate(hz_a):
+                    # solenoid is 1-indexed on hardware
+                    cmd[pump * 3 + solenoid] = f"{hz}|{pump}|{solenoid+1}"
+            elif any(pump - i < 3 for i in self.ipp_pumps):
+                raise ValueError(f"pump slot {pump} reserved for IPP pump, cannot address as standard")
+            else:
+                # TODO: calibration transform here. The time_to_pump is the target variable
+                # where pump_interval would presumably be what the pump_interval was set to
+                # during calibration and is fixed post-calibration.
```
jamienoss, line 62: We need to discuss this for implications impacting #44. I don't think this will lean itself to a clean and extensible 
 calibration framework if we need to scatter the calibration in an ad-hoc manner depending on the device code.

Ideally, the calibration/transform needs to have happened before this gets called. However, the above might also add to the merit of separating the semantics of calibration from something like unit transformation.

jamienoss (changes requested):
One or two minor revisions, a bunch of questions, comments, and a few suggestions to refactor some exiting code that can be punted to another PR.

Also, we really need to discuss any impact that merging this PR may impose on #44 - because, rather than have that dealt with in #86 we may need to instead address it here and alter the pump design/implementation.

evolver/hardware/standard/pump.py
amitschang: The `model_post_init` is that of `Config` if I understand you correctly. Yeah I can add it there. Could you elaborate on the second part? Here I take None to represent that we want the default arrangement so a user doesn't have to explicitly map - the UI could likewise indicate that no setting was explicitly given by the user.

evolver/hardware/standard/pump.py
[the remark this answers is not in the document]
amitschang: Will add there. I don't think it is necessary to be symmetric, it is documented that to set same rate simply either set the same rates or set only influx rate. We of course could simply require both to be explicitly set.

To raise or lower volume the user would set different rates for influx/efflux which is fully supported.

evolver/hardware/standard/stir.py
```diff
@@ -0,0 +1,29 @@
...
+    class Input(VialBaseModel):
+        rate: int = Field(0, description="Stir rate setting")
+
+    @property
+    def serial(self):
+        return self.serial_conn or self.evolver.serial
+
+    def commit(self):
+        inputs = copy(self.committed)
+        inputs.update({v: i for v, i in self.proposal.items() if v in self.vials})
```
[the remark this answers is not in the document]
amitschang: indeed, I will do so in another PR along with the serial property as you suggested above - I've created https://github.com/ssec-jhu/evolver-ng/issues/103 to do that

evolver/hardware/standard/pump.py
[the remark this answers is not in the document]
jamienoss: > We of course could simply require both to be explicitly set.

I'd prefer this option for simplicity, but we'd obviously have to ask the PIs since it changes current and expected behavior.

>To raise or lower volume the user would set different rates for influx/efflux which is fully supported.

True, true, though it would be a better user experience to be able to do this directly by turning one off, rather than having to compute deltas.

evolver/hardware/standard/pump.py
jamienoss: Do you think we may need an endpoint to create model instances for the UI to hit and have the complete model instance returned? If so, it would be better for us to be able to instantiate a ``Config`` model instance and have it be complete with defaults and not have to instantiate the container class to achieve this. In some scenarios we really may not want to instantiate the container class due to other unwanted side-effects, e.g., opening device connections etc. This aspect is also just true for SDK "correctness" - the ``Config`` model should be responsible for its default, not the container (at least were possible).

jamienoss: Well, this is just the exact same thing as above but from the other side, you equally can't instantiate a ``Config`` because it wouldn't be processing the default case.

However, we've discussed this pattern in the past and you decided that you wanted to not duplicate param semantics, defaults, and init functionality across both the container and its ``Config``. The solution is to use the creation pattern we settled on in #73 ( I think that's the correct PR) and call ``class.create()`` which first instantiates a ``Config`` instance and then passes that to ``class.__init__``.

evolver/hardware/standard/pump.py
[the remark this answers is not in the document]
amitschang: I'll make explicit required, it is not too much a burden on user. However I won't here deal with the deltas since there is some complexity as to timing which I don't think we have control over and probably have to go back to PIs to see what use case there is. Can loop back to it later

evolver/hardware/standard/pump.py
amitschang: Specifically I mean here: https://github.com/ssec-jhu/evolver-ng/blob/main/evolver/hardware/test_utils.py#L117 in the tests we instantiate through several available means. If I move the default setting out to Config, the one at that line fails (but I can indeed instantiate a Config since `None` is an allowed value, which is represented in another test).

I'm not sure what to do here - shall I do the repetition of default settings, or get rid of the direct instantiation test (and thereby assume that these should be created with `create` instead)?

jamienoss: I'd go with the latter and use ``create``.

That being said, there's already enough complexity in [evolver.base.BaseInterface.auto_assign_attrs_from_config](https://github.com/ssec-jhu/evolver-ng/blob/main/evolver/base.py#L266) that there could be some argument made for seeing if ``Config.model_post_init`` can also be accommodated. On the other hand, us having to incorporate/replicate ``pydantic`` functionality was my main reason for not wanting to add ``auto_assign_attrs_from_config`` in the first place - so 🤷

jamienoss (approved):

evolver/hardware/standard/pump.py
```diff
@@ -0,0 +1,112 @@
...
+    arduino fluidics module. Pumps are indexed by pump ID (either position in array or
+    IPP pump number) and not vial, each must be controlled separately.
+
+    Specify the index of any IPP pumps via the config parameter ipp_pumps, this should
+    be an array of those pump IDs that are IPP. Please note that IPP pumps reserve 3
+    slots (1 for each solenoid), so if pump 0 is IPP, 1 and 2 cannot be assigned as
+    non-IPP pumps.
+    """
+
+    class Config(SerialDeviceConfigBase):
```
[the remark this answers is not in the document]
jamienoss: @amitschang I missed this yesterday when I approved, but don't see this having been addressed?

jamienoss (changes requested):

jamienoss:
@amitschang Thanks for the update, but it looks like the same issue is true for some of the parent classes.

Take a look at #104, please. I think it would be good to address that first to ensure that this PR is working as expected.

amitschang:
This PR I think is working as expected, even if some of that should to be refactored, so seems simpler to me to merge this in and then changes affecting the `Config` inheritence can work across the board on mainline.

jamienoss (approved):

PR #141, opened 2024-08-02
Add default config

### LINKED ISSUE #140: Problem running app according to README Example Run
Tried to run the application locally using the command `tox -e test exec -- python -m evolver.app.main`.  The following error pops up:

> ERROR:    Traceback (most recent call last):
  File "/Users/ezirayimerwolle/Documents/GitHub.tmp/Khalil_Lab/evolver-ng/.tox/test/lib/python3.11/site-packages/starlette/routing.py", line 732, in lifespan
    async with self.lifespan_context(app) as maybe_state:
  File "/Users/ezirayimerwolle/.pyenv/versions/3.11.9/lib/python3.11/contextlib.py", line 210, in __aenter__
    return await anext(self.gen)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ezirayimerwolle/Documents/GitHub.tmp/Khalil_Lab/evolver-ng/evolver/app/main.py", line 21, in lifespan
    app.state.evolver = Evolver.create(Evolver.Config.load(app_settings.CONFIG_FILE))
                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ezirayimerwolle/Documents/GitHub.tmp/Khalil_Lab/evolver-ng/evolver/base.py", line 75, in load
    with open(file_path, encoding=encoding) as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'evolver.yml'

This issue only started recently, so I imagine there is a recent commit that either broke some sort of logic or the file is no longer being checked in.

jamienoss:
Resolves #140

### DIFF
--- a/evolver.yml
+++ b/evolver.yml
@@ -1,0 +1,34 @@
+controllers: []
+enable_commit: true
+enable_control: true
+hardware: {}
+history:
+  classinfo: evolver.history.HistoryServer
+  config:
+    name: HistoryServer
+interval: 20
+name: Evolver
+serial:
+  classinfo: evolver.serial.EvolverSerialUART
+  config:
+    baudrate: 9600
+    name: EvolverSerialUART
+    port: /dev/ttyAMA0
+    timeout: 1.0
+vials:
+- 0
+- 1
+- 2
+- 3
+- 4
+- 5
+- 6
+- 7
+- 8
+- 9
+- 10
+- 11
+- 12
+- 13
+- 14
+- 15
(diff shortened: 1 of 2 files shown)


jamienoss:
@amitschang thoughts? Should I go ahead here and add some of the standard hardware too?

jamienoss:
We'd prefer to not have to update this with most PRs during this dev stage, so we're going with #149 instead of this PR.

PR #149, opened 2024-08-13
correct port in Readme and do not require evolver.yml by default

imaitland:
i just had a shot at starting `evolver-ng` from scratch, hit same error as Ezira (refs #140 )

I also got the:  `FileNotFoundError: [Errno 2] No such file or directory: 'evolver.yml'` error. 

To fix - I toggled the flag: https://github.com/ssec-jhu/evolver-ng/blob/main/evolver/settings.py#L20. to False as @jamienoss  suggested.

Then i went to the front end i was able to edit and save the configuration to a sane state...


evolver/settings.py
```diff
@@ -23,7 +23,7 @@ class Settings(BaseSettings):
 
 class AppSettings(BaseSettings):
     CONFIG_FILE: Path = Path("evolver.yml")  # in current directory
-    LOAD_FROM_CONFIG_ON_STARTUP: bool = True
+    LOAD_FROM_CONFIG_ON_STARTUP: bool = False
```
amitschang, line 26: I think here `False` is probably not a good idea, since this will need to be flipped to utilize any config that does get written (and typically that is desired). The more immediate fix would be to either instruct the user to copy a stub-out, or instruct them to start the first time as something like

```
EVOLVER LOAD_FROM_CONFIG_ON_STARTUP=false python -m evolver.app.main
```

then next time without. There is a deeper fix to edit the behavior altogether, but scope here could probably be just the startup docs as-is

jamienoss: I second this, setting this to ``False`` for this reason doesn't seem correct as they'd only want to turn it back on after bootstrapping. I think this should be ``True`` by default.

jamienoss: With this change, this pr is also in conflict/dupe of #141

amitschang (dismissed):
looks good, thanks!

README.md
```diff
@@ -37,8 +37,7 @@ see Build instructions below ❗❗
 
 ### Config file
 
-Configuration of the eVolver system including provisioned hardware and experiments can be expressed in a single yaml
-file, for example:
+Configuration of the eVolver system including provisioned hardware and experiments must be expressed in a single yaml, `evolver.yml` stored in the project root directory file, for example:
```
jamienoss, line 40: This isn't strictly true as the filename is custom by changing [evolver.settings.AppSettings.CONFIG_FILE](https://github.com/ssec-jhu/evolver-ng/blob/main/evolver/settings.py#L25).

The env var ``EVOLVER_CONFIG_FILE`` can also be used.

I think we should note this in the docs.

jamienoss: It obviously doesn't even have to be in the project root dir.

README.md
```diff
@@ -63,6 +62,24 @@ controllers:
 This enables both sharing of the eVolver setup and experiment with others, and also the ability to easily resume the
 experiment on hardware failure.
 
+#### Bootstrapping the `evolver.yml` config file
+
+By default `evolver-ng` requires an `evolver.yml` file to exist in the project's root directory for it to start.
```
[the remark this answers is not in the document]
amitschang: or iso `"some_filename.yml"` use `settings.*` (this should be whatever the default file name is)

README.md
imaitland: updated the readme to make this clear.

README.md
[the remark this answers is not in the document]
imaitland: updated the readme to make this clear

jamienoss (approved):
👍 
Just a minor suggestion for you to do as you please with... and an annoying comment about the port that should have been noted from the beginning 😆 sorry 😉

jamienoss (approved):

jamienoss:
BTW I was able to figure out why the codecov test was not running since we upgraded.

Should we:
 1) make @imaitland admin so that they can admin merge their PRs?
 2) unmark that test as required?

imaitland:
> BTW I was unable to figure out why the codecov test was not running since we upgraded.
> 
> Should we:
> 
>     1. make @imaitland admin so that they can admin merge their PRs?
> 
>     2. unmark that test as required?

Probably want to keep codecov as a blocker... make me admin.

amitschang (approved):

amitschang:
I'll go ahead and merge this in

PR #154, opened 2024-08-26
add id

imaitland:
Add an id field to the evolver and return it in the `describe` response.

This gives the frontend a stable identifier for the evolver(s) it has connected to.

### DIFF
--- a/evolver/device.py
+++ b/evolver/device.py
@@ -1,4 +1,5 @@
 import time
+import uuid
 from collections import defaultdict
 
 from evolver.base import BaseInterface, ConfigDescriptor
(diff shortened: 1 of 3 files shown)


evolver/device.py
```diff
@@ -28,6 +29,7 @@ class Config(BaseInterface.Config):
         interval: int = settings.DEFAULT_LOOP_INTERVAL
 
     def __init__(self, *args, **kwargs):
+        self.id = uuid.uuid4().hex[:6]
```
jamienoss, line 32: There might be some scope for using this as a tracer amongst logs, history, & data, etc. Perhaps at least for sanity of which box produced what. In which case, this might be better off in the ``Config``.

jamienoss: Just saw @amitschang slcak comment:
>Technically maybe the mac address of the underlying network device (e.g. on the rPi) might more stably identify the particular device?

I think this is a great idea as it gives users a way of physically knowing which box (hardware) produced stuff (as long as it also gets dumped out with the config and everything else).

PR #155, opened 2024-08-26
fix chemostat for pump, test against real hw Input model

amitschang:
This fixes the flow_rate settings for pumping, which require influx and efflux to be explicitly set. 

It also adds to the test a check against the input model. I think later we can pull out some of this mocking infra to create utilities that help with mocking hardware interaction (set, get, Input, Output) for controllers against references like this, but leaving this isolated in this chemostat for current change.


evolver/controller/standard/chemostat.py
```diff
@@ -80,5 +80,5 @@ def control(self, *args, **kwargs):
             # Inputs assume the relevant device has its calibration and takes
             # the target real value. These may be missing spec, for example does
             # pump need bolus value also?
-            self.pump.set(self.pump.Input(vial=vial, flow_rate=self.flow_rate))
+            self.pump.set(self.pump.Input(vial=vial, flow_rate_influx=self.flow_rate, flow_rate_efflux=self.flow_rate))
```
jamienoss, line 83: I don't think this works without breaking some reusability, right? ``chemostat`` has only semantics for a single ``flow_rate``, and no boundary on what ``self.pump`` can be. If the pump is a ``GenericPump`` this will break as its ``Input`` has no ``flow_rate_influx`` and ``flow_rate_efflux``. This suggests that we're either not using our interface correctly here or we need to improve the pump interface.

jamienoss: TBH, this highlights an issue with the interface around the use of the ``Input`` class within control code as it's not agnostic to hardware when used like this. I think we'd have to make ``set`` take a dict and then have the hardware process it into its ``Input``, i.e., ``self.pump.set(flow_rate=self.flow_rate``, where the base ``set`` is ``set(**kwargs): input = self.Input(**kwargs); ...`` and can be overridden by ``VialIEPump``  to account for a single ``flow_rate``. Or we have to create an adequate interface for ``Input`` on each hardware type, which may be needed anyhow, and might also be the better design option. It would be better if each hardware "kind" adhered to a single interface which can then be swapped around via config without breaking or needing to change control code. Even with this, we may still find it better to _also_ have ``set`` not take an instance of ``Input``, though if we have an adequate interface it shouldn't be necessary.

jamienoss (changes requested):
This looks to break some hardware interfaces and/or highlights that we're missing an interface aspect/layer.

jamienoss:
Regression from #100

evolver/controller/standard/chemostat.py
amitschang: Won't this work for any pump who takes the parameters `vial`, `flow_rate_influx`, and `flow_rate_efflux`? This is the requirement and it should only be expected to work with pumps that present that interface (which genericpump does not - vialiepump is composed of 2 of them per vial).

I think it might be a nicer ergonomics to have `set` (or another setter, like `set_from`) that takes kw args, but that wouldn't change the requirements on the hardware. If we want `VialIEPump` to accept a single flow rate, then its input should change, but I think we opted for explicit treatment in its PR.

Chemostat works with influx and efflux pump to move material in/out, so this doesn't break anything wrt that.

jamienoss: >Chemostat works with influx and efflux pump to move material in/out, so this doesn't break anything wrt that.

This is abstract though, and is easily imagined to take a single flow rate if desired and easily imagined that the hardware used can also just take a single flow rate and it's the hardware code's responsibility to figure that out given the specific physics/setup of the hardware being used. Otherwise, we defeat the purpose of creating layers of abstraction.

>Won't this work for any pump who takes the parameters vial, flow_rate_influx, and flow_rate_efflux?

Of course, but as you point out, not for the ``GenericPump`` which looked to me to be the interface class for pumps, so should be adhered to. Adding these two attrs to the interface is an option perhaps alongside ``flow_rate`` 🤷 . We should think about this some more.

amitschang:
I've updated the VialIEPump interface to accept a single flow_rate that is set across influx/efflux, mutually exclusive with setting the individual rates

jamienoss (approved):
Cool, thanks!

PR #157, opened 2024-08-29
147 call transformers for temp

### LINKED ISSUE #147: Call transformers for std hardware: temp

jamienoss:
Resolves #147

### DIFF
--- a/evolver/calibration/standard/polyfit.py
+++ b/evolver/calibration/standard/polyfit.py
@@ -5,6 +5,7 @@ from pydantic import Field, model_validator
 
 from evolver.base import ConfigDescriptor
 from evolver.calibration.interface import Calibrator, Transformer
+from evolver.settings import settings
 
 
 class PolyFitTransformer(Transformer):
@@ -50,12 +51,13 @@ class LinearTransformer(PolyFitTransformer):
 
 
 class PolyFitCalibrator(Calibrator):
-    def run_calibration_procedure(self, data: dict):
+    def run_calibration_procedure(self, data: dict, save=settings.AUTO_SAVE_NEW_CALIBRATIONS):
         # TODO: This is just a placeholder.
         calibration_data = {}
         for transformer, (x, y) in data.items():
             calibration_data[transformer] = getattr(self, transformer).refit(x, y)
-            calibration_data[transformer].save()
+            if save:
+                calibration_data[transformer].save()
         return calibration_data
 
 
@@ -63,3 +65,32 @@ class LinearCalibrator(PolyFitCalibrator):
     class Config(PolyFitCalibrator.Config):
         input_transformer: ConfigDescriptor | LinearTransformer | None = None
         output_transformer: ConfigDescriptor | LinearTransformer | None = None
+
+
+class IndependentVialBasedLinearCalibrator(PolyFitCalibrator):
+    class Config(PolyFitCalibrator.Config):
+        """Specify transformers for each vial independently. Whilst they may all use the same transformer class, each
+        vial will mostly likely have different transformer config parameters and thus require their own transformer
+        instance.
+        """
+
+        input_transformer: dict[int, ConfigDescriptor | LinearTransformer | None] | None = None
+        output_transformer: dict[int, ConfigDescriptor | LinearTransformer | None] | None = None
+
+    def run_calibration_procedure(self, data: dict):
+        """Override to implement a calibration procedure that is vial-dependant and thus calibrates all vials either
+        simultaneously or entirely as desired."""
+
+        calibration_data = dict(input_transformer={}, output_transformer={})
+
+        # Sequentially run calibration procedure for the input transformer for each vial.
+        for vial, input_transformer in self.input_transformer.items():
+            calibration_data["input_transformer"][vial] = input_transformer.run_calibration_procedure(
+                data["input_transformer"][vial]
+            )
+
+        # Sequentially run calibration procedure for the output transformer for each vial.
+        for vial, output_transformer in self.output_transformer.items():
+            calibration_data["output_transformer"][vial] = output_transformer.run_calibration_procedure(
+                data["output_transformer"][vial]
+            )
--- a/evolver/calibration/standard/tests/test_fitting.py
+++ b/evolver/calibration/standard/tests/test_fitting.py
@@ -4,7 +4,10 @@ import numpy as np
 import pytest
 from pydantic import ValidationError
 
+import evolver.settings
+from evolver.base import ConfigDescriptor
 from evolver.calibration.standard.polyfit import LinearCalibrator, LinearTransformer
+from evolver.tests.conftest import tmp_calibration_dir  # noqa: F401
 
 
 @pytest.fixture(scope="class")
@@ -59,7 +62,7 @@ class TestLinearTransformer:
 
 
 class TestLinearCalibrator:
-    def test_calibration_procedure(self, mock_linear_data):
+    def test_calibration_procedure(self, mock_linear_data, tmp_calibration_dir):  # noqa: F811
         x, y, c = mock_linear_data
         obj = LinearCalibrator(input_transformer=LinearTransformer(coefficients=c))
         old_config = obj.input_transformer.config_model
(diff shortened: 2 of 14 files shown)


evolver/calibration/standard/tests/test_fitting.py
```diff
@@ -4,9 +4,16 @@
...
 from pydantic import ValidationError
 
+import evolver.settings
+from evolver.base import ConfigDescriptor
 from evolver.calibration.standard.polyfit import LinearCalibrator, LinearTransformer
 
 
+@pytest.fixture()
+def tmp_calibration_dir(monkeypatch, tmp_path):
+    monkeypatch.setattr(evolver.settings.settings, "ROOT_CALIBRATOR_FILE_STORAGE_PATH", tmp_path)
```
jamienoss, line 14: Semi-unrelated, but I noticed that the tests were creating persisting calibration files.

evolver/hardware/standard/temperature.py
```diff
@@ -21,13 +24,16 @@ class Temperature(SensorDriver, EffectorDriver):
...
     class Config(SerialDeviceConfigBase, EffectorDriver.Config):
-        pass
+        calibrator: ConfigDescriptor | Calibrator | None = Field(
+            default_factory=lambda: ConfigDescriptor.load(settings.DEFAULT_TEMPERATURE_CALIBRATION_CONFIG_FILE),
+            description="The calibrator used to both calibrate and transform Input and/or Output data.",
+        )
 
     class Output(SerialDeviceOutputBase, SensorDriver.Output):
-        temperature: float = Field(None, description="Sensor temperature in degrees celcius")
+        temperature: float = Field(None, description="Sensor temperature in degrees Celsius")
```
jamienoss, line 33: smuggle

jamienoss:
@amitschang Hey, do you mind taking a preliminary look at this, please?

Main design aspect is that the vial transformers are ``{vial: Transformer()}`` instead of the vial list/dict being a attr of a specific "multi-vial" transformer.

evolver/calibration/standard/polyfit.py
```diff
@@ -55,11 +55,40 @@ def run_calibration_procedure(self, data: dict):
...
+    class Config(PolyFitCalibrator.Config):
+        """Specify transformers for each vial independently. Whilst they may all use the same transformer class, each
+        vial will mostly likely have different transformer config parameters and thus require their own transformer
+        instance.
+        """
+
+        input_transformer: dict[int, ConfigDescriptor | LinearTransformer | None] | None = None
+        output_transformer: dict[int, ConfigDescriptor | LinearTransformer | None] | None = None
+
+    def run_calibration_procedure(self, data: dict):
```
jamienoss, line 78: FYI this is just a place-holder.

evolver/calibration/standard/polyfit.py
```diff
@@ -55,11 +55,40 @@ def run_calibration_procedure(self, data: dict):
...
+
+class IndependentVialBasedLinearCalibrator(PolyFitCalibrator):
+    class Config(PolyFitCalibrator.Config):
+        """Specify transformers for each vial independently. Whilst they may all use the same transformer class, each
+        vial will mostly likely have different transformer config parameters and thus require their own transformer
+        instance.
+        """
+
+        input_transformer: dict[int, ConfigDescriptor | LinearTransformer | None] | None = None
+        output_transformer: dict[int, ConfigDescriptor | LinearTransformer | None] | None = None
```
jamienoss, line 76: Main feature.

evolver/calibration/standard/polyfit.py
```diff
@@ -50,16 +51,46 @@ def convert_from(self, y):
...
 class PolyFitCalibrator(Calibrator):
-    def run_calibration_procedure(self, data: dict):
+    def run_calibration_procedure(self, data: dict, save=settings.AUTO_SAVE_NEW_CALIBRATIONS):
         # TODO: This is just a placeholder.
         calibration_data = {}
         for transformer, (x, y) in data.items():
             calibration_data[transformer] = getattr(self, transformer).refit(x, y)
-            calibration_data[transformer].save()
+            if save:
+                calibration_data[transformer].save()
```
jamienoss, line 60: smuggle

amitschang:
Seems good, Just a couple impressions:

* I was imagining that the calibration procedure would be most flexible if it stored (or additionally had the option to store) the points of `raw` mapping to `real`, since then one could swap out fit functions, but here the coefficients from the fit look to be what is in the config file, if I'm reading it right. Any comment?
* In the temperature driver the code allows both global and vial, but I'm not sure there is a good scope for global coefficients so might be simpler to just require per-vial there.

jamienoss:
@amitschang thanks for the prelim review.

>In the temperature driver the code allows both global and vial, but I'm not sure there is a good scope for global coefficients so might be simpler to just require per-vial there.

Can you point to the code, as I don't see the ref to global coeffs.

> I was imagining that the calibration procedure would be most flexible if it stored (or additionally had the option to store) the points of raw mapping to real, since then one could swap out fit functions, but here the coefficients from the fit look to be what is in the config file, if I'm reading it right. Any comment?

That's a good idea. I think it would have to do both though, allow the ``calibrator`` to store the raw data (``measuredData`` as they ref to it) which could be null but also allow for the coeffs to be explicitly passed to the transformers - along with some logic/flags to account for the presence of both and exceptions etc. The only issue here is that I don't have any ``measuredData `` for temp, just the coeffs in https://raw.githubusercontent.com/FYNCH-BIO/evolver/master/evolver/calibrations.json. But that can be added later I guess.

evolver/hardware/standard/temperature.py
```diff
@@ -54,8 +60,16 @@ def read(self):
...
-                self.outputs[vial] = self.Output(vial=vial, raw=int(raw))
+                # Calibrate raw data to temperature.
+                raw = int(raw)
+                temperature = None
+                if self.calibrator and self.calibrator.output_transformer:
+                    if isinstance(self.calibrator.output_transformer, dict):
+                        temperature = self.calibrator.output_transformer[vial].convert_to(raw)
+                    else:
+                        temperature = self.calibrator.output_transformer.convert_to(raw)
+
```
amitschang, line 71: Sorry @jamienoss , I should have just put it here. I mean that maybe we don't need to bother with the non-vial case, the procedure really should be producing one per vial even if there is only one vial

jamienoss: Cool, thanks. I re-read your comment and realised what you were referring to.

I thought it was more useful and extensible to allow for both. I've also factored this out to the hardware interface as:
```python
    def _transform(self, transformer, func, x, vial=None):
        if self.calibrator and (_transformer := getattr(self.calibrator, transformer, None)):
            if isinstance(_transformer, dict):
                y = getattr(_transformer[vial], func)(x)
            else:
                y = getattr(_transformer, func)(x)
        else:
            y = x

        return y
```

amitschang: 👍 ah, ok this will simplify while also being flexible

jamienoss:
@amitschang FYI I'm gonna punt ``Calibrator.Config.measured_data`` (or its similar implementation) to #165. This way I can handle the issue of the coeffs being required by the transformers later and get all of the other hardware transformers merged instead.

evolver/hardware/interface.py
```diff
@@ -25,6 +26,18 @@ def __init__(self, *args, evolver=None, **kwargs):
         self.evolver = evolver
         super().__init__(*args, **kwargs)
 
+    def _transform(self, transformer: str, func: str, x: Any, vial: int = None):
```
amitschang, line 29: I wonder if it might be more general if we don't call the `func` but rather return a transformer, and provide a signature more like `transformer_in(self, vial) -> Transformer` (effectors) and `transformer_out(self, vial) -> Transformer` (sensors), called like:

```py
temp = self.transformer_out(vial).convert_to(raw)
```

it can be reused a bit easer since one could stash the transformer to make further calls if needed. It reads slightly more intentional (arguably)

evolver/tests/conftest.py
```diff
@@ -1,6 +1,13 @@
...
+from evolver.settings import settings
+
 
 @pytest.fixture
 def testme():
     return "value"
+
+
+@pytest.fixture
+def tmp_calibration_dir(monkeypatch, tmp_path):
```
amitschang, line 12: might be useful as an autouse?

amitschang (commented):
Looks good, just a couple small comments.

evolver/hardware/interface.py
jamienoss: Could do, but I didn't necessarily want to add it to the hardware interface, which the above suggestion would. There'd then need to be two new funcs: ``transformer_in`` and ``transformer_out`` (which I think would be better suited as properties), however, this would still leave most of the boilerplate as they could both return ``None`` and the caller would still have to check for this before calling, e.g., ``transformer.convert_to``, and handle the noop.  So it would read:
```python
if transformer:=self.transform_in(vial):
  temp = transformer.convert_to(raw)
else:
  temp = raw
```
or
```python
temp = self.transformer_in(vial).convert_to(raw) if self.transformer_in(vial) else raw
```
Thoughts?

evolver/tests/conftest.py
jamienoss: Actually, this doesn't work as the fixture order matters.

amitschang (approved):
sounds good, thanks!

PR #158, opened 2024-08-30
Iainland/healthz html response

imaitland:
Screenshot 2024-08-30 at 12 06 20 PM

As part of the work to make the on-board 7" display show useful information about the evolver, we need an endpoint that responds with useful html.

This can be fleshed out as the requirements for what the on-board display should show emerge.


evolver/app/main.py
```diff
@@ -81,9 +83,24 @@ async def get_history(
     )
 
 
-@app.get("/healthz", operation_id="healthcheck")
+@app.get("/healthz", operation_id="healthcheck", response_class=HTMLResponse)
 async def healthz():
-    return {"message": f"Running '{__project__}' ver: '{__version__}'"}
+    hostname = socket.gethostname()
+    ip_address = socket.gethostbyname(hostname)
```
jamienoss, line 89: We should consider #164

jamienoss (approved):
👍 

Not for this PR but I opened #163 as it would be cool to report the on/off status of the 12V system in the healthcheck.

jamienoss (approved):
👍

PR #169, opened 2024-09-12
Call transformers for OD

### LINKED ISSUE #144: Call transformers for std hardware: od_sensor

jamienoss:
Resolves #144

### DIFF
--- a/evolver/hardware/standard/od_sensor.py
+++ b/evolver/hardware/standard/od_sensor.py
@@ -1,4 +1,5 @@
 from pydantic import Field
+from pydantic.fields import FieldInfo
 
 from evolver.base import ConfigDescriptor
 from evolver.calibration.interface import Calibrator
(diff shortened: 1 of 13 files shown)


jamienoss:
@jamienoss refactor linearfit, etc, ``coefficients`` -> ``parameters`` to make uniform with other transformers.

evolver/hardware/standard/od_sensor.py
```diff
@@ -29,5 +32,26 @@ def read(self):
...
+                # Calibrate raw data to density.
+                raw = int(raw)
+                density = self._transform("output_transformer", "convert_to", raw, vial)
+                self.outputs[vial] = self.Output(vial=vial, raw=raw, density=density)
         return self.outputs
+
+
+class OD90(ODSensor):
+    class Config(ODSensor.Config):
+        name: str = ODSensor.Config.model_fields["name"].from_field(default="od_90")
```
amitschang, line 44: I'm not sure this is correct - I think the intention is to combine the `Field` object information, essentially overriding the default? but this looks like it doesn't do that (`from_field` is a static method)

```py
f = pydantic.Field(default='a', description='test')
f.from_field(default='b')
-> FieldInfo(annotation=NoneType, required=False, default='b')
```

I'm not sure the way this is supposed to be done, but there is a merge field info:
```py
FieldInfo.merge_field_infos(f, FieldInfo(default='b'))
-> FieldInfo(annotation=NoneType, required=False, default='b', description='test')
```

amitschang (commented):
Some comments

evolver/hardware/standard/od_sensor.py
jamienoss: Nice catch, I hadn't noticed that, thanks!

jamienoss: Rereading the [the docs for this](https://docs.pydantic.dev/latest/api/fields/#pydantic.fields.FieldInfo.from_field), I really don't see the point of this method at all. Seems like quite a wasted opportunity and my, assumed functionality, would be far more useful. Never mind...

jamienoss: ``merge_field_infos`` did the trick, thanks again.

amitschang (approved):
looks good, thanks!

PR #174, opened 2024-09-26
starting on pump transform, but thinking about calibration generally too

amitschang:
This adds calibrator for the pumps, at the same time as adds a facility for those to store data in a named "calibration file" which can be used to store user data to refit upon load. This facility is added in the interface so can be reused. I'm not sure with this how to utilize a user-configured transformer (we want a class, not an instance for that), but probably can approach that next.

Also, I had the pump operation wrong, they don't control primarily in flow rate but in injections of volume (and optionally a frequency by which to do the injections). The PR unfortunately includes a bit of distraction refactoring that, sorry 😄

### DIFF
--- a/evolver/calibration/interface.py
+++ b/evolver/calibration/interface.py
@@ -125,6 +125,7 @@ class Calibrator(BaseInterface):
     class Config(Transformer.Config):
         input_transformer: ConfigDescriptor | Transformer | None = None
         output_transformer: ConfigDescriptor | Transformer | None = None
+        calibration_file: str | None = None
 
     class CalibrationData(Transformer.Config): ...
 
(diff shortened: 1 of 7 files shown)


imaitland (approved):
LGTM

evolver/calibration/interface.py
```diff
@@ -125,6 +125,9 @@ class state(_BaseConfig):
     class Config(Transformer.Config):
         input_transformer: ConfigDescriptor | Transformer | None = None
         output_transformer: ConfigDescriptor | Transformer | None = None
+        calibration_file: str | None = None
```
jamienoss, line 128: @amitschang I've opened # to address this but we had discussed this a few times before and decided that we didn't want to add this as it's not necessary and self-referential.

All configs can be loaded from file using ``model_validate``, the calibration files are no exception here.
See https://github.com/ssec-jhu/evolver-ng/blob/main/evolver/base.py#L52-L53
```python
class _BaseConfig(pydantic.BaseModel):
    model_config = pydantic.ConfigDict(extra="ignore", from_attributes=True)

    @classmethod
    def model_validate(cls, obj, *, strict=None, from_attributes=None, context=None):
        """Effectively the same as pydantic.BaseModel.model_validate() except that it automatically handles json, and
        conversion from instances of ``BaseConfig`` and ``BaseInterface`
        """
        if obj is None:
            return cls()
        elif isinstance(obj, os.PathLike):
            return cls.load(file_path=obj)
        ...
```

It's quite possible that there are a few code paths that need patching or re-directing to ensure that this works from all nominal invocations, # can contain that if needed.

PR #185, opened 2024-10-10
Save config upon exiting evolver context

jamienoss:


evolver/device.py
```diff
@@ -92,3 +92,12 @@ def loop_once(self):
             self.evaluate_controllers()
         if self.enable_commit:
             self.commit_proposals()
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.config_model.save(app_settings.CONFIG_FILE)
+
```
jamienoss, line 101: We could also add here...
```suggestion
  self.enable_control = False
  self.enable_commit = False
```

evolver/app/main.py
```diff
@@ -28,8 +28,11 @@ async def lifespan(app: FastAPI):
...
         app.state.evolver = Evolver.create()
-    asyncio.create_task(evolver_async_loop())
-    yield
+
+    with app.state.evolver:
+        asyncio.create_task(evolver_async_loop())
+        yield
+
     # Shutdown:
     ...
```
jamienoss, line 37: Note that such "save" could go here also/instead, however, the more encapsulated place for it to live is in ``Evolver``.

amitschang (commented):
I'm okay with it, but since it doesn't really guarantee that any modified internal state is propogated to the file we can't rely on it and should do `update` any time we want to change the config in a reliable way - I'm afraid it might get confusing

jamienoss:
>but since it doesn't really guarantee that any modified internal state is propogated to the file

It's not supposed to. But that also depends on what you mean by _guarantee_. This will _always_ write the config if the app exits (as long as it's not from a signal interrupt or python interp crash etc).

However, if you also mean a guarantee that modified state gets dumped to file when modified, the ``update`` endpoint doesn't do this either as state can be modified outside of the update, i.e., within control/experiment code where the user is free to do so.

I think what you might be expecting is periodic snapshots, however, this is something additional, and not something that this PR is supposed to address.

jamienoss:
>should do update any time we want to change the config in a reliable way

Yep, that's exactly what the ``update`` endpoint is for, this is not supposed to reflect any functionality of mutating the config.

amitschang (approved):

PR #186, opened 2024-10-10
small fix to tox check style to match what format would change

amitschang:
There was a mismatch in `check-style` and `format`, where `check-style` passed while format changed some things even in passing case. This means that some things could (and did) get missed which create confusing changes for subsequent PRs.

Only one line is relevent, `tox.ini:17`, everything else is a fix for the style as a result


imaitland (approved):
LGTM.

PR #187, opened 2024-10-10
Handling exceptions in loop for application run

### LINKED ISSUE #152: Address exception handling in Evolver loop activity
The loop activity contains the critical function experiment logic, we should both ensure that exceptions become visible to the user and that exceptions are handled in a manner appropriate to the device function (e.g. abort, continue experiment, etc)

amitschang:
In the running application we want to keep the system running on best effort, so the loop should not raise a fatal exception due to conditions of the hardware or controllers. The proposed method reads/executes/commits as much as possible while logging issues, but another method would be to fail on first exception and capture that in the loop. There is config option to control behavior.

resolves #152


jamienoss:
Sorry for prying whilst this is a draft, however, I really don't think we want to do this, at least not as proposed.

Normally, you add one outermost exception handler for logging, we can add two due to the app/SDK nature. So we'd need exception handling in the app (already handled for us) and then I think the other best place would be in ``evolver. device.Evolver.loop_once``. That would then only need to look like the following:
```
def loop_once():
  try:
    ...
  except Exception:
    self.logger.exception(f"Error in loop:")
    raise
```
I don't think we ever want the need to keep limping on if an error occured, this would be very misleading, a warning would at least need to be raised, and even then it could be detrimental to experimental results as the logs would always need to be checked to determine whether results are valid.

Again, sorry for commenting whilst this is only a draft.

evolver/device.py
```diff
@@ -71,19 +72,27 @@ def schema(self):
...
+            callable()
+        except Exception as e:
+            if self.raise_loop_exceptions:
+                raise e
+            self.logger.error(f"Error in loop: {message}: {e}")
+
     def read_state(self):
         for name, device in self.sensors.items():
-            device.read()
+            self._loop_exception_wrapper(device.read, f"reading device {name}")
```
jamienoss, line 85: Wouldn't this imply that the caller would be ok with possibly receiving garbage? This could trigger a down stream exception that would only raise anyhow.

evolver/device.py
```diff
@@ -71,19 +72,27 @@ def schema(self):
...
 
     def evaluate_controllers(self):
         for controller in self.controllers:
-            controller.run()
+            self._loop_exception_wrapper(controller.run, f"updating controller {controller}")
 
     def commit_proposals(self):
         for device in self.effectors.values():
-            device.commit()
+            self._loop_exception_wrapper(device.commit, f"committing proposals for {device}")
```
jamienoss, line 95: Masking exceptions here could be considered unsafe as the physical state of the device will be unknown.

amitschang:
I thought about wrapping at the higher level, but I decided to do it this way so that if there is some transient issue with say the serial line on a single hardware we can log the error(which would be made available as warning to user) and keep trying others. After an exception in any of these we don't know necessarily that the system is all bad, next loop could get back to good functioning.

For the commit, that is true, what we could do is allow the user to decide if they want to stop committing actions (pause) or even abort on errors in the previous actions, which would require a manual action to recover. They already have the option to just raise errors and not skip (which would presently cause the system to do cease to function). Alternatively or in addition we could try all reads but have the option to exit the loop if any of them raised an exception.

jamienoss:
>After an exception in any of these we don't know necessarily that the system is all bad, next loop could get back to good functioning.

This is true, it could, however, the difficulty is how would you know that your experimental results weren't impacted by this?

evolver/device.py
```diff
@@ -71,19 +72,27 @@ def schema(self):
             "controllers": [{"kind": str(type(a)), "config": a.Config.model_json_schema()} for a in self.controllers],
         }
 
+    def _loop_exception_wrapper(self, callable, message="uknown"):
+        try:
+            callable()
+        except Exception as e:
+            if self.raise_loop_exceptions:
+                raise e
+            self.logger.error(f"Error in loop: {message}: {e}")
```
jamienoss, line 81: Would it not be better to always log here?
```suggestion
            self.logger.error(f"Error in loop: {message}: {e}")
            if self.raise_loop_exceptions:
                raise e
```

jamienoss: Also, ``logger.exception`` would be better as it includes the traceback.

evolver/device.py
amitschang: maybe, but if they do receive something, that something would have been added to output by the driver and it would be responsible for either delivering garbage or not, I think. And yeah, if the consumer of it raised that would be handled and logged in the controller section.

amitschang:
Since this is arleady being reviewed, I will open it up. 

How about this latest change? The user has the option to skip the rest of control loop on read error, which is probably the safest thing (so defaults to true). If things come back then the controllers can implement there logic based on latest state. 

One reason to handle individually is that some transient error in a controller should not necessarily prevent another from keeping its experiment going just cause it was after in the list. The experiment conditions can possibly be recovered in subsequent runs by experiment logic - the real danger would be things like accidentally pumping too much and overflowing -  but not sure there is a good handle on that since no direct volume measurement, if this is a risk, user could choose to abort on failure (we can handle this when we handle stopping conditions).

amitschang:
**Note**:  We are going to ask the BU folks what are the desired behaviors in case of failure at read/logic/commit phases, before proceeding further with this PR.

amitschang:
Spoke with Ezira. They currently ignore transient errors in the serial comms from rPI to hardware (analogous to our `read` step), and have exception handling for experiment code where the evolver will stop (analogous to our `control` step and `abort`). They agreed that matching this behavior with the various options like to abort on control exception is desired as an escape hatch, but that generally an experiment can go on with low risk.

I've updated this PR with the option `abort_on_control_errors` - together with other options here I believe captures the desired behaviors.

evolver/device.py
```diff
@@ -25,6 +25,9 @@ class Config(BaseInterface.Config):
         history: ConfigDescriptor | History = ConfigDescriptor.model_validate(DEFAULT_HISTORY)
         enable_control: bool = True
         interval: int = settings.DEFAULT_LOOP_INTERVAL
+        raise_loop_exceptions: bool = False
+        abort_on_control_errors: bool = False
+        skip_control_on_read_failure: bool = True
```
jamienoss, line 30: I don't think this gives us what you were looking for. Looks like it's missing ``abort_on_commit _errors`` or similarly named.

As we have it here, to raise commit errors ``raise_loop_exceptions`` needs to be ``True``. However, this is only checked in ``_loop_exception_wrapper`` and therefore takes precedence over both ``abort_on_control_errors`` & ``skip_control_on_read_failure``. This means the scenario where you don't raise on read and control errors but do on commit isn't covered, right?

I would have thought raising on commit errors to be the most important - assuming we make any distinction at all.

evolver/device.py
```diff
@@ -70,25 +73,43 @@ def schema(self):
             "controllers": [{"kind": str(type(a)), "config": a.Config.model_json_schema()} for a in self.controllers],
         }
 
+    def _loop_exception_wrapper(self, callable, message="unknown") -> bool:
+        try:
+            callable()
```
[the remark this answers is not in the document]
amitschang: I wont do this for now if it is ok. It'll make meesage require to be keyword in the call, and since it is marked private I think scope is limited enough

evolver/device.py
```diff
@@ -70,26 +74,49 @@ def schema(self):
...
+        if read_error and self.skip_control_on_read_failure:
+            self.logger.info("Skipping control loop due to read error")
+            return
         if self.enable_control:
-            self.evaluate_controllers()
-            self.commit_proposals()
+            control_error = self.evaluate_controllers()
+            if control_error and self.abort_on_control_errors:
+                self.abort()
+                return
```
jamienoss, line 115: @amitschang Sorry, I should have noted this before - what's the rationale for not wanting to actually ``raise`` here? I would have expected this to raise the error not just ``return``?

amitschang: Since we captured the errors, logged and aborted, we can assume we are in a "clean" state (from app perspective). If the user were to re-enable the flags cause they deem it OK, then the application could continue.

We can raise here, but then would need to capture it in the webapp loop activity, but then I think we might need to differentiate between errors raised by `_raise_loop_exceptions` and others.

evolver/device.py
jamienoss: 😮

evolver/device.py
jamienoss: Sure, but the web app is its own responsibility. However, when using this programmatically, we would expect it to raise, no? I feel like the alternative is wiping python's exception handling functionality clean off the map.

amitschang: OK. I guess we should then raise an exception if any captured error happened within the loop, not just if the thing was aborted.

jamienoss: I might be wrong, but I think that only hangs on the semantics of "abort" being name-mangled into the flags, right? Otherwise, I'm not following your point and feel like I missing something for that to be true.

Ignoring the "abort" name-mangle for a mo, I see the flags as more of a "continue-don't-raise" (inverted bool on what we have here). So to me, the ``raise`` is _the_ point here, however, we gracefully "abort" because an exception was raised so that hardware state can be "reset".  The "abort" isn't the focus from my perspective.

Again, as noted before, this screams for just using context-managers to handle the abort and thus keeping the focus of these flags and conditional checks on the option of whether or not to raise etc.

amitschang: what do you mean by "name-mangled"? 

I might have misinterpreted what you said before then. Do you mean that because we choose to abort on the erorr, it should raise some exception, not that it should raise to indicate any error was encounted. How would it be in the case of skip-on-read? Or strictly for abort since it is like a termination condition?

amitschang: something like latest change?

jamienoss: RE name-mangling: ``abort_on_control_errors`` & ``abort_on_commit_errors``.

>How would it be in the case of skip-on-read?

Skip-on-read is ok, because we're explicitly skipping and continuing.

>Or strictly for abort since it is like a termination condition?

Exactly, the ``abort`` terminates (the app can deal with its own continuation semantics/functionality) however, these flags are in the ``Evolver.Config`` and should be agnostic to the app.

My main concern here is that if we abort and just return and continue at this layer, for these three core functions, aren't we nuking all exceptions from being raised (unless ``raise_loop_exceptions==True``). I think this is the point that we should focus on, if I'm wrong then we should be ok. But it seems to me that this PR does away with exceptions being raised, since the default for ``raise_loop_exceptions`` is ``False`` with core focus being shifted to the use of the other three flags which absorb raised exceptions.

Perhaps, my misunderstanding is with the following, I see these flags as completely overlapping and not different functionality/semantics, but _just_ granularity. I.e., where this def is true: raise_loop_exceptions := abort_on_control_errors & abort_on_commit_errors & ~skip_control_on_read_failure.

evolver/device.py
```diff
@@ -70,26 +74,49 @@ def schema(self):
...
+        if read_error and self.skip_control_on_read_failure:
+            self.logger.info("Skipping control loop due to read error")
+            return
         if self.enable_control:
-            self.evaluate_controllers()
-            self.commit_proposals()
+            control_error = self.evaluate_controllers()
+            if control_error and self.abort_on_control_errors:
+                self.abort()
+                raise RuntimeError("Aborted due to control error")
```
jamienoss, line 115: This raises _an_ error but not _the_ error(s) and it would be cool if we could chain them. If the funcs returned a list of errors rather than just a bool, we could use ``from``, e.g., ``raise RuntimeError("Aborted due to control error") from control_error[0]`` (or chain them all - though the 1st might be enough, with a note that there could be more).

amitschang: I think runtime error would be fine. I can do the exception from thing, I'll keep it simple and just raise from first one and say see logs.

jamienoss (approved):
Awesome! Thanks! 👍

PR #189, opened 2024-10-11
Iainland/calibration actions

imaitland:
This PR:
- adds a `UserInput` pydantic model to the calibration actions
- refactors the `TemperatureCalibrator` into `calibration/standard/calibrators/temperature.py`
- parses the `UserInput` in the dispatch method of CalibrationProcedure.

misc cleanup
- the list actions endpoint can be a `get`

### DIFF
--- a/evolver/calibration/actions.py
+++ b/evolver/calibration/actions.py
@@ -2,23 +2,37 @@ from abc import ABC, abstractmethod
 from copy import deepcopy
 from typing import Any, Dict
 
+import pydantic
+
 
 class CalibrationAction(ABC):
+    class UserInput(pydantic.BaseModel): ...
+
     @abstractmethod
     def execute(self, state: Dict[str, Any], payload: Dict[str, Any] = None) -> Dict[str, Any]:
         pass
 
 
 class DisplayInstructionAction(CalibrationAction):
+    class UserInput(pydantic.BaseModel):
+        confirmation: bool = pydantic.Field(..., description="Confirm that the instruction has been completed")
+
     def __init__(self, description: str, name: str):
         self.description = description
         self.name = name
 
     def execute(self, state: Dict[str, Any], payload: Dict[str, Any] = None) -> Dict[str, Any]:
+        user_input = self.UserInput(**payload)
+        if not user_input.confirmation:
+            raise ValueError("Confirmation is required to proceed")
+
         return state.copy()
 
 
 class VialTempReferenceValueAction(CalibrationAction):
+    class UserInput(pydantic.BaseModel):
+        temp: float = pydantic.Field(..., title="Temperature", description="Temperature in degrees Celsius")
+
     def __init__(self, hardware, description: str, vial_idx: int, name: str):
         self.hardware = hardware
         self.description = description
--- a/evolver/calibration/actions.py
+++ b/evolver/calibration/actions.py
@@ -9,21 +9,20 @@ class CalibrationAction(ABC):
     class UserInput(pydantic.BaseModel): ...
 
     @abstractmethod
-    def execute(self, state: Dict[str, Any], payload: Dict[str, Any] = None) -> Dict[str, Any]:
+    def execute(self, state: Dict[str, Any], payload: UserInput) -> Dict[str, Any]:
         pass
 
 
 class DisplayInstructionAction(CalibrationAction):
     class UserInput(pydantic.BaseModel):
-        confirmation: bool = pydantic.Field(..., description="Confirm that the instruction has been completed")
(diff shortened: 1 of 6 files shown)


evolver/calibration/actions.py
```diff
@@ -2,23 +2,37 @@
...
+    class UserInput(pydantic.BaseModel): ...
+
     @abstractmethod
     def execute(self, state: Dict[str, Any], payload: Dict[str, Any] = None) -> Dict[str, Any]:
         pass
 
 
 class DisplayInstructionAction(CalibrationAction):
+    class UserInput(pydantic.BaseModel):
+        confirmation: bool = pydantic.Field(..., description="Confirm that the instruction has been completed")
```
amitschang, line 18: for this boolean case we can probably default it to true, so the logic below (`if not confirmation`) would be activated instead of a de-serialization error

evolver/calibration/actions.py
```diff
@@ -2,23 +2,37 @@
...
+        user_input = self.UserInput(**payload)
+        if not user_input.confirmation:
+            raise ValueError("Confirmation is required to proceed")
+
         return state.copy()
 
 
 class VialTempReferenceValueAction(CalibrationAction):
+    class UserInput(pydantic.BaseModel):
+        temp: float = pydantic.Field(..., title="Temperature", description="Temperature in degrees Celsius")
```
amitschang, line 34: optional, but I don't think the `...` is necessary

evolver/calibration/actions.py
```diff
@@ -2,23 +2,37 @@
...
 class DisplayInstructionAction(CalibrationAction):
+    class UserInput(pydantic.BaseModel):
+        confirmation: bool = pydantic.Field(..., description="Confirm that the instruction has been completed")
+
     def __init__(self, description: str, name: str):
         self.description = description
         self.name = name
 
     def execute(self, state: Dict[str, Any], payload: Dict[str, Any] = None) -> Dict[str, Any]:
+        user_input = self.UserInput(**payload)
```
amitschang, line 25: I wonder if the input type should actually be the already deserialized `UserInput`, this could be done upper layers by the framework (maybe in `calibrationProcedure.dispatch` , since then it knows the action)? In other words the API becomes:

```py
def execute(self, state: Dict[str, Any], payload: UserInput = None) -> ...
```

amitschang (commented):
Looks mostly good, just a couple comments

evolver/calibration/actions.py
imaitland: ah got it, nice, will change.

evolver/calibration/actions.py
imaitland: will fix.

evolver/calibration/actions.py
amitschang: sorry, I meant default to `false`!!

evolver/calibration/actions.py
imaitland: Nice idea, will implement.

evolver/calibration/actions.py
```diff
@@ -2,34 +2,44 @@
...
+
     @abstractmethod
-    def execute(self, state: Dict[str, Any], payload: Dict[str, Any] = None) -> Dict[str, Any]:
+    def execute(self, state: Dict[str, Any], payload: UserInput) -> Dict[str, Any]:
         pass
 
 
 class DisplayInstructionAction(CalibrationAction):
+    class UserInput(pydantic.BaseModel):
+        confirmation: bool = pydantic.Field(True, description="Confirm that the instruction has been completed")
```
amitschang, line 18: small one, my fault - I think default should probably be `False`, forcing user to answer. Though now I wonder if this action even benefits from that? To get here they would have to press a button or something?

imaitland: Just a checkbox, so if you instruct the user to wait for environment to equilibriate for example, they have to click a checkbox confirming they did that step...

imaitland: and yeah, duh default of True makes no sense will fix.

imaitland: actually on reflection i think this is unecessary friction, the user can be trusted to follow instructions, and the confirmation checkbox  doesn't guarantee it anyway.

imaitland: yeah False is the correct default but really this isn't really useful for an instruction step. I was thinking it'd be good to force the user to confirm the instruction was followed but i think it's just friction and will be irritating.

amitschang (approved):
lgtm, thanks for the updates!

PR #190, opened 2024-10-14
[history] Add in-memory buffer of configurable number of partitions

### LINKED ISSUE #183: Resolve latency issue in history endpoint
The standard history implementation is file-backed, which has a lot of nice features but even with the time-based partitioning it is noticeably slow on the raspberry pi.

Maybe storing a certain window in-memory, finding a format more suitable, or maybe just tuning partition size, etc might help. 

subtasks:
- [ ] make json reading more efficient by not requiring sampling (see #184)
- [ ] Keep in memory buffer

amitschang:
This is second of set of changes to make history perform better. For recent data (which should be most common to query) we additionaly store in-memory and query that if possible. Data are always also written out to disk and queryable when not satisfied in memory buffer.

brings UI latency from a couple seconds down to 10s of milliseconds for default query.

Resolves #183

### DIFF
--- a/evolver/history/standard.py
+++ b/evolver/history/standard.py
@@ -55,8 +55,10 @@ class HistoryServer(History):
                 f"SELECT * FROM {self.json_hist_reader} WHERE time_part>=?",  # nosec: B608
                 params=(start_part,),
             )
-        except duckdb.IOException:
-            return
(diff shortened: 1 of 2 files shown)


imaitland (approved):
LGTM.

PR #191, opened 2024-10-14
enable setting hardware inputs from web api

### LINKED ISSUE #181: Enable Hardware Control from API
This essentially means setting the desired state on an Effector (`Effector.set`), but via web API call instead of via controller. Noting that we should probably either:

* Enforce that the control loop be turned off since the command may quickly get overridden
* Warn that the command may quickly get overridden - may be useful when a controller doesn't specifically control given hardware

amitschang:
This change adds endpoints to set hardware inputs from the web api. Adds also some exception handling related to these changes. Along with a smuggled fix for temp inputs.

resolves #181 

**Note**: This allows one to set a value even when the controllers are enabled and might just overwrite it (not guaranteed to be overwritten, just possible). I *think* this is OK here, and any warning should probably be implemented at client level (UI or CLI client), where user interaction is managed.


imaitland (approved):
LGTM

PR #159, opened 2024-08-30
controller dev docs wip

amitschang:

PR #192, opened 2024-10-15
Concepts doc

amitschang:
Initial version of the concepts doc that is linked elsewhere. It references a number of classes/methods so adding the API docs to make those links work as well.

### DIFF
--- a/docs/source/concepts.rst
+++ b/docs/source/concepts.rst
@@ -76,6 +76,8 @@ As related to the above yaml configuration. We can see the top-level class
 ``ODSensor`` represented in ``classinfo`` and the config values as modeled in
 the ``Config`` class represented within the ``config`` key.
 
+.. _experiment_loop:
+
 Experiment loop
 ---------------
 
@@ -144,6 +146,12 @@ penalty, simplifying the controller code, for example when looping over the set
 of configured vials.
 
 .. note::
+  Note that values are typically neither read nor committed within the `control`
+  method itself - these are executed by the evolver in the **read** and
+  **commit** phases of :ref:`experiment_loop`. Multiple ``get`` calls would
+  return the same values and subsequent ``set`` calls would overwrite the value
+  to commit.
+
   While technically it is feasible to call the `read` and `commit` methods
   within a controller, we recommend against doing so. Due to the serial
   communication latency, and the fact that a commit will be done for all
--- a/docs/source/concepts.rst
+++ b/docs/source/concepts.rst
@@ -1,4 +1,60 @@
 Evolver concepts
 ================
 
-TBD
+The evolver system is made up of several configurable components tied together
+by a manager component that orchestrates the application control loop and
+exposure to the web interface. Components can be grouped into two high level
+categories:
+
+* **Hardware and experiment control components**: These are the main target of extensions,
+  representing the physical devices and experiment logic reading from and
+  controlling them.
+
+  * :py:class:`HardwareDriver<evolver.hardware.interface.HardwareDriver>`, specifically sensors that can be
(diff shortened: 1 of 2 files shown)


imaitland (dismissed):
Lgtm

docs/source/concepts.rst
```diff
@@ -1,4 +1,158 @@
...
+
+1. **Read sensors**: values are `read` and buffered within individual hardware
+   drivers, for all configured sensors. For more on reading sensors and
+   buffering see :ref:`buffering`.
+
+2. **Execute controllers**: the `control` method of controllers is called for
+   all controllers, in sequence. The controllers can `get` sensor values that
+   have previously been read, and `set` effector values.
+
+3. **Commit effector values**: effector values are `commit`ted for all
```
jamienoss, line 93:
```suggestion
3. **Commit effector values**: effector values are `committed` for all
```

docs/source/concepts.rst
```diff
@@ -1,4 +1,158 @@
...
+  the read phase of the current loop.
+* writes of effector values should be done using the
+  :py:meth:`set<evolver.hardware.interface.EffectorDriver.set>` method of the
+  `EffectorDriver` interface, which will buffer the value to be committed in the
+  commit phase of the current loop.
+
+These methods can be called as many times as required with no additional
+penalty, simplifying the controller code, for example when looping over the set
+of configured vials.
+
```
jamienoss, line 145:
```suggestion
.. note::
  Remember that values are neither read nor commanded within the `control` method itself - these are executed
  within the main evolver loop :meth:`evolver.device.Evolver.loop_once`. Consequently, multiple sequential reads
  and commits are not possible within a single call to :meth:`evolver.controller.Controller.control`. The physical state of the hardware cannot be altered more than once within this method - and technically it is not altered at all until after this method returns.
```

jamienoss: I still think it is a good idea for us to really emphasize and drive this point home.
As commented before in https://github.com/ssec-jhu/evolver-ng/pull/159#discussion_r1752251535.

amitschang: Do you have a suggestion for how to do so beyond having a section for it here? This would be available for linking in other places in the documentation to keep users aware

jamienoss: Nope.

amitschang: I added a note to drive point home, bit different from above

docs/source/concepts.rst
```diff
@@ -1,4 +1,165 @@
...
+.. note::
+  While technically it is feasible to call the `read` and `commit` methods
```
jamienoss, line 147:
```suggestion
  Remember that values are neither read nor commanded within the `control` method itself - these are executed
  within the main evolver loop :meth:`evolver.device.Evolver.loop_once`. Consequently, multiple sequential reads
  and commits are not possible within a single call to :meth:`evolver.controller.Controller.control`. While technically it is feasible to resolve this by calling the `read` and `commit` methods
```
Your comment as is, still doesn't express nor warn against the problem, that someone adding multiple ``get``s are redundant and multiple ``set``s will clobber themselves.

amitschang: the above is not quite true though, as it mentions it is not possible. It is just not the intended design. Maybe I should add some extra language that calling get multiple times gives the same result (though I think the whole section covers this...)

amitschang: reads a bit confusing to me. Does the latest update get at it, you think? (I also don't want to frame it as a problem that needs to be resolved)

jamienoss (approved):

amitschang:
thanks!

PR #200, opened 2024-10-28
adding post-abort start and active check endpoints

### LINKED ISSUE #198: Add a start endpoint to start/restart control after having aborted
Relates to #194

amitschang:
Resolves #198

Does so by adding a start endpoint which is basically just an update setting `enable_control` and `enable_commit` to true. It also adds an "active" flag, exposed both in healthz and state, which indicates if those activities are being done.


evolver/device.py
```diff
@@ -71,6 +71,10 @@ def schema(self):
             "controllers": [{"kind": str(type(a)), "config": a.Config.model_json_schema()} for a in self.controllers],
         }
 
+    @property
+    def active(self):
+        return self.enable_control and self.enable_commit
```
jamienoss, line 76: I'm just wondering how this might intuitively get interpreted by someone manually controlling the hardware, i.e., not using ``control`` but committing commands?

"active" sounds to me like they should be able to but the above implies that you shouldn't/can't. We could have a "running" status flag, however at that point, you literally just have these two flags. Perhaps it's best to nuke this state reduction and just natively report both ``enable_control`` and ``enable_commit`` in the ``healthz`` resp?

amitschang: Really lack of `commit` is what should make it inactive, but no control would generally mean nothing to `commit`.

But also I wonder if maybe these needn't be separate. Perhaps I can just make `enable_control` mean both execute control and commit - the separation is probably not too useful...

jamienoss: >Really lack of commit is what should make it inactive, but no control would generally mean nothing to commit.

Exactly.

>But also I wonder if maybe these needn't be separate. Perhaps I can just make enable_control mean both execute control and commit - the separation is probably not too useful...

Do you mean separate at all or just as reported from ``healthz``? 

With the way we have the configuration setup, I def think it's needed to have both. One deactivates hardware control (``enable_commit``) and the other starts the controller(s)/experiment running (``enable_control``). Removing ``enable_control`` would require someone who wants to pause an experiment (control) and manually interact with the hardware to have to stop the running experiment by creating a new config with the ``control`` section omitted. Seems burdensome to me, no?

jamienoss: >>Really lack of commit is what should make it inactive, but no control would generally mean nothing to >commit.
>
>Exactly.

Actually, I'm not sure if that's entirely accurate as ``enable_commit`` is only checked in ``evolver.device.Evolver.loop_once`` and not at any communication layer. So a user could programmatically control the hardware still from the SDK (but not the app).

jamienoss (approved):

PR #206, opened 2024-11-11
Iainland/refactor history endpoint - split GET and POST handlers

imaitland:
HTTP standards demand GET requests do not include a body, all GET params must be passed as query elements.


imaitland:
ok got it, yeah we should have only one. I'm ok with POST only. Will make that change.

using GET and a body was causing the TS client to complain. https://github.com/whatwg/fetch/issues/551

amitschang (approved):

PR #210, opened 2024-11-18
make config file, working dir agnostic, enables debugging of evolver.app.main module in pycharm

imaitland:
Working Pycharm debug on `evolver.app.main` module. Useful for (frontend) development against the http api.

Screenshot of working config.

Screenshot 2024-11-18 at 3 21 50 PM

### DIFF
--- a/evolver/settings.py
+++ b/evolver/settings.py
@@ -34,7 +34,9 @@ class Settings(BaseSettings):
 
 
 class AppSettings(BaseSettings):
-    CONFIG_FILE: Path = Path("evolver.yml")  # in current directory
+    CONFIG_FILE: Path = (
+        Path(__file__).resolve().parent.parent / "evolver.yml"
+    )  # CONFIG_FILE is in the project root, regardless of the working directory.
     LOAD_FROM_CONFIG_ON_STARTUP: bool = True
     HOST: str = "127.0.0.1"
     PORT: int = 8080
--- a/evolver/settings.py
+++ b/evolver/settings.py
@@ -34,12 +34,26 @@ class Settings(BaseSettings):
 
 
 class AppSettings(BaseSettings):
-    CONFIG_FILE: Path = (
-        Path(__file__).resolve().parent.parent / "evolver.yml"
-    )  # CONFIG_FILE is in the project root, regardless of the working directory.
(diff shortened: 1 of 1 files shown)


evolver/settings.py
```diff
@@ -34,11 +34,27 @@ class Settings(BaseSettings):
 
 
 class AppSettings(BaseSettings):
-    CONFIG_FILE: Path = Path("evolver.yml")  # in current directory
+    def find_project_root(self, marker: str = "pyproject.toml") -> Path:
```
jamienoss, line 37: So this already exists [evolver.util.find_package_location](https://github.com/ssec-jhu/evolver-ng/blob/main/evolver/util.py#L10).

evolver/settings.py
```diff
@@ -34,11 +34,27 @@ class Settings(BaseSettings):
...
+
+    CONFIG_FILE: Path | None = None
     LOAD_FROM_CONFIG_ON_STARTUP: bool = True
     HOST: str = "127.0.0.1"
     PORT: int = 8080
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if self.CONFIG_FILE is None:
+            self.CONFIG_FILE = self.find_project_root() / "evolver.yml"
```
jamienoss, line 56: This is already what the app will do, kinda, assuming that it's started from the repo dir.

jamienoss (commented):
All you have to do is edit the config file with an absolute path or rel to the dir you want for ``CONFIG_FILE`` or set the env var ``EVOLVER_CONFIG_FILE`` before spinning up the application. Either wat I don't think this change is needed.

evolver/settings.py
imaitland: yeah but not if I am using a debugger and start from a module, e.g. evolver.app.main, in that case the working dir for the process is 2 dirs deep relative to the project root, so it can't find the evolver.yml.

evolver/settings.py
imaitland: ah nice.

evolver/settings.py
jamienoss: You should change the working dir in your debugger

PR #214, opened 2024-11-21
Always serialize configs as ConfigDescriptors

### LINKED ISSUE #213: Bug in nested rebuilding of Config as ConfigDescriptor
```python

#calibrator=None
print(x)
```

This could be due to:
 * changes in ``pydantic``
 
jamienoss:
Resolves #213

### DIFF
--- a/evolver/device.py
+++ b/evolver/device.py
@@ -19,7 +19,7 @@ class Evolver(BaseInterface):
         name: str = "Evolver"
         experiment: str = "unspecified"
         vials: list = list(range(settings.DEFAULT_NUMBER_OF_VIALS_PER_BOX))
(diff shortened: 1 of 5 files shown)


evolver/device.py
```diff
@@ -19,7 +19,7 @@ class Config(BaseInterface.Config):
         name: str = "Evolver"
         experiment: str = "unspecified"
         vials: list = list(range(settings.DEFAULT_NUMBER_OF_VIALS_PER_BOX))
-        hardware: dict[str, ConfigDescriptor | HardwareDriver] = {}
```
jamienoss, line 22: We shouldn't have to do this, nor do we want to, but it does resolve the issue.

jamienoss:
Replaced by #216

PR #216, opened 2024-11-23
Add nested config test and temp downgrade of pydantic

### LINKED ISSUE #213: Bug in nested rebuilding of Config as ConfigDescriptor

jamienoss:
Relates to #213

This still needs further investigation to determine whether something will remain broken for newer versions of ``pydantic``.  This does not resolve #213, but will temporarily band-aid it.

### DIFF
--- a/evolver/tests/test_base.py
+++ b/evolver/tests/test_base.py
@@ -5,6 +5,9 @@ import pytest
 
 import evolver.base
 import evolver.util
+from evolver.calibration.standard.calibrators.temperature import TemperatureCalibrator
+from evolver.device import Evolver
+from evolver.hardware.standard.temperature import Temperature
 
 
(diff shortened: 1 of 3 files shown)


evolver/tests/test_base.py
```diff
@@ -219,6 +222,22 @@ def test_from_file(self, mock_config_file):
         assert obj.a == 44
         assert obj.b == 55
 
+    def test_nested_config_models(self):
```
jamienoss, line 225: This test fails on anything greater than ``pydantic=2.7.4``.

jamienoss:
@amitschang ``<2.8`` of shall we tether it to exactly what we know works, i.e., ``==2.7.4``?

amitschang (approved):
Looks good

PR #217, opened 2024-11-23
Refactor evolver.base

jamienoss:
#213 Was the main driver for this refactor. However, it does not fix that issue but instead refactors some code to make things more manageable to future changes.

The core change is the creation of ``_BaseModel`` for the separation of conversion back and forth between ``Config``, instances, and ``ConfigDescriptors`` such that it is clearer where in the code this happens - it is no longer circular. For clarity ``_BaseModel`` does not handle the conversion between ``Config`` and ``configDescriptor`` and instead only includes other base conversions, i.e., json strings. Soln. to #234 would also go in this new class. This PR introduces some duplication in logic, but at the expense of it being more direct and explicit per specific desired class behavior.

This should also help and relate to #234.

### DIFF
--- a/evolver/base.py
+++ b/evolver/base.py
@@ -77,29 +77,6 @@ class _BaseModel(pydantic.BaseModel):
         """
         return dict(self)
 
-
-class _BaseConfig(_BaseModel):
-
-    @classmethod
-    def get_classinfo(cls):
-        """The fully qualified classname for cls's container class (if one exists).
-           Raises `TypeError` if parent class is not derived from BaseInterface.
-        """
-        fqn = evolver.util.fully_qualified_name(cls)
-        containers = fqn.split('.')[:-1]
-        container = ".".join(containers)
-        container_class = import_string(container)
-        if issubclass(container_class, BaseInterface):
-            cls = container_class
-        return f"{cls.__module__}.{cls.__qualname__}"
-
-    # @pydantic.model_serializer(mode="wrap", when_used='json')
-    # def to_descriptor(self, handler) -> dict:
-    #     serialized_by_super = handler(self)
-    #     if isinstance(serialized_by_super, dict) and (set(serialized_by_super.keys()) == {"classinfo", "config"}):
-    #         return serialized_by_super
-    #     return dict(classinfo=self.get_classinfo(), config=serialized_by_super)
-
     @classmethod
     def model_validate(cls, obj, *, strict=None, from_attributes=None, context=None):
         """Effectively the same as pydantic.BaseModel.model_validate() except that it automatically handles json, and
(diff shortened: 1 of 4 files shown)


evolver/base.py
```diff
@@ -92,6 +75,75 @@ def shallow_model_dump(self):
...
+                # Ok, maybe it wasn't a ConfigDescriptor... (or was but was malformed)
+                # Explicitly call ``model_validate_json`` rather than ``super().model_validate()`` to declare explicit
+                # code logic.
+                return cls.model_validate_json(obj, strict=strict, context=context)
+            else:
+                # Cool, it was a ConfigDescriptor so validate from that. Remember ``ConfigDescriptor.config`` is just a
+                # dict.
+                return super().model_validate(
+                    descriptor.config, strict=strict, from_attributes=from_attributes, context=context
+                )
```
jamienoss, line 126: @amitschang It would be handy for #234 if this duplication got nuked. Does it make sense for us to lose info by converting from a descriptor to a naked ``Config``?

It appears that there's currently no test in the suite for this (which I can add), however, I can't think of an existing of needed use case for this, can you?

jamienoss: Test added.

amitschang: Do you mean a need for converting from a `ConfigDescriptor` to `Config` (of the class specified in `classinfo`, I guess)? If the loss you are talking about is when serializing that `Config` object we lose the `classinfo`, then yeah I can't think of a reason we desire this.

But not sure if that is what you are asking.

jamienoss: Yeah, that's exactly it.

amitschang: @jamienoss, is there something you will change here, then? Or is it ready?

jamienoss: Not in this PR as it opens questions to removing other sections of that switch also, i.e., all other sections dealing with descriptors, not just those as strings.

Sorry for the confusion, I just wanted to know as an aid to working on #234.

jamienoss: Opened #242 to address this.

evolver/base.py
```diff
@@ -8,6 +8,7 @@
 import pydantic
 import pydantic_core
 import yaml
+from pydantic._internal._validators import import_string
```
amitschang, line 11: Should we be using this? the `_internal` and lack of docs make me think not. It looks like this is used to import from fqcn to a type, perhaps we could either make a util using importlib or a simple wrapper around a pydantic model that has ImportString type?

evolver/base.py
```diff
@@ -215,35 +265,45 @@ class BaseInterface(ABC):
...
             try:
                 descriptor = ConfigDescriptor.model_validate(config, context=dict(extra="forbid"))
-                config = validate_descriptor(descriptor)
             except pydantic.ValidationError:
                 config = cls.Config.model_validate(config)
+            else:
+                validate_classinfo(descriptor.classinfo)
+                config = descriptor.config
+
+        config_dict = config.shallow_model_dump() if isinstance(config, _BaseConfig) else config
```
amitschang, line 303: It might be worth a note here for posterity as to why this is needed - is it because this could be a dict due to 287 or 301? In the latter case, should it be validated or has it already been?

amitschang (commented):
Looks good overall, just a couple comments

evolver/base.py
jamienoss: Comment added. Also, note that this should be checked against ``_BaseModel`` and not ``_BaseConfig``.

jamienoss: ...and it will have been validated by then.

evolver/base.py
```diff
@@ -215,35 +264,47 @@ class BaseInterface(ABC):
...
+            config = cls.Config.model_validate(config["config"])
+        elif isinstance(config, dict):
+            config = cls.Config.model_validate(config)
         else:
-            # Next handle dict | str, which could be that representing a Config or a ConfigDescriptor.
-            # First see if config is actually a descriptor by trying to create one, since it's only fields are limited
-            # to classinfo and config.
+            # Last is a str representing either a Config or ConfigDescriptor.
             try:
                 descriptor = ConfigDescriptor.model_validate(config, context=dict(extra="forbid"))
```
jamienoss, line 295: Re https://github.com/ssec-jhu/evolver-ng/pull/217#discussion_r1910563789, validation takes place here.

evolver/base.py
```diff
@@ -215,35 +264,47 @@ class BaseInterface(ABC):
...
+        elif isinstance(config, _BaseConfig):
+            pass  # Already a config instances so do nothing.
         elif isinstance(config, ConfigDescriptor):
-            config = validate_descriptor(config)
+            validate_classinfo(config.classinfo)
+            config = config.config  # This has already been validated but is still just a dict.
+        elif _is_descriptor_dict(config):
+            validate_classinfo(config["classinfo"])
+            config = cls.Config.model_validate(config["config"])
+        elif isinstance(config, dict):
```
jamienoss, line 290: Re https://github.com/ssec-jhu/evolver-ng/pull/217#discussion_r1910563789, validation takes place here.

amitschang (approved):

Issue #131, opened 2024-08-01
Add endpoint to retrieve logs


jamienoss:
Static dump of logs for the past X unit of time - mimic the API for the history, that way you can namespace the logs.

Issue #132, opened 2024-08-01
Add ui to view logs


imaitland:
refs, #131 

Maybe users can make a slack reference implementation but ui needs basic view into evolver log.

Issue #109, opened 2024-07-31
Proposal: add config update diff to history


imaitland:

imaitland:
This ui task to overlay log data (exceptions, config updates, other events) with the history time series data.

amitschang:
We got a request from Nate which I think might be nice to build in from the start:

Classify (e.g. draw with different colors/styles) the vertical lines by some property. At the moment we don't handle all extra data, but #220 should fix that. Then we can perhaps have a known property or enable selection of property to classify by.

amitschang:
See https://github.com/ssec-jhu/evolver-ui/pull/18 for initial version adding line in

PR #218, opened 2024-11-25
Hookup for log/event messages to history server

### LINKED ISSUE #196: Record abort events
Relates to #109, #194 

Currently, #194 only halts control and commit but not read. When reviewing past collected data, this maybe misleading to someone analyzing the collected data. To remedy this, an abort event delta/tag should be added to the data to indicate that all data beyond this event is read-only and not controlled.

amitschang:
To utilize the existing history server for logs in addition to sensor data, the following steps were taken:
* namespace all included loggers with the package name. This is done so we can modify the logger without affecting root
* add a log handler class that emits to the history API
* add a hookup between the above two within the device manager, along with config options
* modify the history api to take a "kind" and "vial" parameter - kind differentiates between sensor reads and logs, and vial enables logs to be tied to a specific vial
* change shape of sensor data to be row-per-vial, instead of row containing struct of vials (will require corresponding UI mod)

Resolves: #131, Resolves #196
Enables/relates to: #109, #132

Some examples:

```sh
curl -Hcontent-type:application/json -XPOST localhost:8080/history/'?t_start=1732570420&t_stop=1732570440' -d'{"kinds": ["event", "sensor"], "vials": [0, 1]}' | jq .
```
```json
{
  "data": {
    "test": [
      {
        "timestamp": 1732570421.0073779,
        "kind": "sensor",
        "vial": 0,
        "data": {
          "name": "NoOpSensorDriver",
          "vial": 0,
          "raw": 1,
          "value": 2
        }
      },
      {
        "timestamp": 1732570421.009811,
        "kind": "sensor",
        "vial": 1,
        "data": {
          "name": "NoOpSensorDriver",
          "vial": 1,
          "raw": 1,
          "value": 2
        }
      }
    ]
  }
}
```

```sh
curl -Hcontent-type:application/json -XPOST localhost:8080/history/'?n_max=2' -d'{"kinds": ["log"]}
' | jq . 
```
```json
{
  "data": {
    "test": [
      {
        "timestamp": 1732569561.055786,
        "kind": "log",
        "vial": null,
        "data": {
          "level": "WARNING",
          "message": "test from HW"
        }
      }
    ],
    "evolver": [
      {
        "timestamp": 1732569561.033202,
        "kind": "log",
        "vial": null,
        "data": {
          "level": "ERROR",
          "message": "Error in loop: committing proposals for <MagicMock spec='EffectorDriver' id='4987471824'>"
        }
      }
    ]
  }
}
```

### DIFF
--- a/evolver/base.py
+++ b/evolver/base.py
@@ -312,7 +312,7 @@ class BaseInterface(ABC):
         init_and_set_vars_from_descriptors(self, **non_config_kwargs)
 
     def _setup_logger(self):
-        self.logger = logging.getLogger(self.name)
+        self.logger = logging.getLogger(f"{__package__}.{self.name}")
 
     @classmethod
     def __get_pydantic_core_schema__(cls, *args, **kwargs):
--- a/evolver/base.py
+++ b/evolver/base.py
@@ -10,6 +10,7 @@ import pydantic_core
 import yaml
 
 import evolver.util
+from evolver.settings import settings
 from evolver.types import CreatedTimestampField, ExpireField, ImportString
 
 
@@ -312,7 +313,7 @@ class BaseInterface(ABC):
         init_and_set_vars_from_descriptors(self, **non_config_kwargs)
 
     def _setup_logger(self):
-        self.logger = logging.getLogger(f"{__package__}.{self.name}")
+        self.logger = logging.getLogger(f"{settings.DEFAULT_LOGGER}.{self.name}")
 
     @classmethod
     def __get_pydantic_core_schema__(cls, *args, **kwargs):
--- a/evolver/logutils.py
+++ b/evolver/logutils.py
@@ -4,7 +4,7 @@ EVENT = logging.INFO + 1
 logging.addLevelName(EVENT, "EVENT")
 
 
-class CaptureHandler(logging.Handler):
+class LogHistoryCaptureHandler(logging.Handler):
     def __init__(self, history_server):
         super().__init__()
         self.history_server = history_server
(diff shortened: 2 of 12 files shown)


evolver/logutils.py
```diff
@@ -0,0 +1,23 @@
...
+class LogHistoryCaptureHandler(logging.Handler):
+    def __init__(self, history_server):
+        super().__init__()
+        self.history_server = history_server
+
+    def emit(self, record):
+        vial = getattr(record, "vial", None)
+        record_dict = {"level": record.levelname, "message": record.getMessage()}
+        # TODO: add support for some extra fields (but these should be json-seralizable)
+        kind = "event" if record.levelno == EVENT else "log"
```
jamienoss, line 16: [comment] It might help the TODO implementer if we added here and now, how we might envision coding this, e.g., using a [``match``](https://docs.python.org/3/tutorial/controlflow.html#match-statements), an ``enum``, a mapping etc?

evolver/base.py
```diff
@@ -312,7 +312,7 @@ def init_descriptors(self, **non_config_kwargs):
         init_and_set_vars_from_descriptors(self, **non_config_kwargs)
 
     def _setup_logger(self):
-        self.logger = logging.getLogger(self.name)
+        self.logger = logging.getLogger(f"{__package__}.{self.name}")
```
jamienoss, line 315: [comment] It might be slightly more robust against future changes if we added a setting for this rather than relying on the convention of it always being ``__package__``.
E.g.:
```python
# evolver.settings
class Settings(...):
  ROOT_LOGGER_NAME: str = __package__
```

jamienoss (approved):
LGTM 👍 

Just some comments/suggestions.

amitschang:
Thanks. I made some corresponding changes, I think scope of them is small enough I will consider approval sticky and merge this in 😄

PR #221, opened 2024-12-06
ignore pydantic for now

amitschang:
As noted, due to https://github.com/ssec-jhu/evolver-ng/issues/213, see also https://github.com/ssec-jhu/evolver-ng/pull/219


jamienoss (approved):

jamienoss:
FYI I've opened #222 so that we don't forget to undo this.

PR #226, opened 2024-12-10
Log pass through user supplied extra data

amitschang:
This change adds user supplied extra fields to the log, via a helper that does namespacing and validation. 

The namespacing is useful since the logger adds all `extra` fields directly to the log record `__dict__`, which ends up making it impossible to find what was added specifically by the user. The helper here adds fields under a specific keyword to keep this context.

The history endpoint also should expect that data are json serializable (in most general cases), so the helper provides facility to do this. 

In general it can be used like:
```py
logger.info('message', extra=LogInfo(vial=1, arbitrary_info='info'))
```
but provides `dump` in case it needs to be spread into other extra data (which spreads fields under the predefined namespace key):
```py
logger.info('message', extra={'not_saved': 1, **LogInfo(vial=1).dump()})
```

### DIFF
--- a/evolver/logutils.py
+++ b/evolver/logutils.py
@@ -10,11 +10,15 @@ logging.addLevelName(EVENT, "EVENT")
 class LogInfo:
     """Helper class to provide "extra" to log records integrated with history.
 
-    This class acts similar to a dict, but with all fields within a dict under a
-    specific key and ensures the fields are json-serializable. Used with the
-    history integration, fields provided will be available in the stored record.
-    This is useful because all fields within extra are placed at the top level
-    of a logrecord dict, so otherwise there is no way to distinguish them.
+    This class acts similar to a dict, but with all fields within it "keyed" by
+    an EXTRA_KEY that is stripped when the log event is read. This has the
+    effect of normalizing LogInfo data, so that when it is read it appears at
+    the same level as the Log entry's top-level properties.
+
+    Note:
+        This is useful because all fields within extra are placed at the top
+        level of a logrecord dict, so otherwise there is no way to distinguish
+        them.
 
     Example:
         self.logger.log(EVENT, "Event message", extra=LogInfo(vial=1))
--- a/evolver/logutils.py
+++ b/evolver/logutils.py
@@ -8,6 +8,19 @@ logging.addLevelName(EVENT, "EVENT")
 
 
 class LogInfo:
+    """Helper class to provide "extra" to log records integrated with history.
+
+    This class acts similar to a dict, but with all fields within a dict under a
+    specific key and ensures the fields are json-serializable. Used with the
+    history integration, fields provided will be available in the stored record.
+    This is useful because all fields within extra are placed at the top level
+    of a logrecord dict, so otherwise there is no way to distinguish them.
+
+    Example:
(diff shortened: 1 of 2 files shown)


evolver/logutils.py
```diff
@@ -6,24 +7,63 @@
...
+
+    @staticmethod
+    def json_check(data, silent=False):
+        if data == {}:
+            return data
+        try:
+            json.dumps(data)
+            return data
+        except TypeError as e:
+            if silent:
```
imaitland, line 34: curious why silent is needed?

evolver/tests/test_logging.py
```diff
@@ -24,14 +26,14 @@ def test_log_evolver_hookup():
     class HW(SensorDriver):
         def read(self):
             self.logger.warning("test from HW")
-            self.logger.log(EVENT, "event from HW", extra={"vial": 1})
+            self.logger.log(EVENT, "event from HW", extra=LogInfo(vial=1, event_class="something"))
```
imaitland, line 29: I am curious - what happens if I pass in extra and it's not an instance of LogInfo? would this part happen:
```
ValueError(f"Extra data must be json-serializable, got error: {e}")
```

evolver/logutils.py
```diff
@@ -6,24 +7,63 @@
...
 
 
+class LogInfo:
+    """Helper class to provide "extra" to log records integrated with history.
+
+    This class acts similar to a dict, but with all fields within a dict under a
+    specific key and ensures the fields are json-serializable. Used with the
+    history integration, fields provided will be available in the stored record.
+    This is useful because all fields within extra are placed at the top level
+    of a logrecord dict, so otherwise there is no way to distinguish them.
```
imaitland, line 17: "data in the LogInfo class is "keyed" by an `EXTRA_KEY` that is stripped when the log event is read. This has the effect of normalizing LogInfo data, so that when it is read it appears at the same level as the Log entry's top-level properties. 

For example: self.logger.log(EVENT, "event from HW", extra=LogInfo(vial=1, event_class="something")) will produce a serialized log entry: {"level": "EVENT", "message": "event from HW", "event_class": "something", "vial": 1}

imaitland (dismissed):
Some questions, but LGTM.

evolver/tests/test_logging.py
amitschang: You could pass in extra alone, but if you wanted it to appear in history you would specifically do:

```
extra={'evolver_extra': {'stuff_for_history': ...}}
```

and you'd have to be sure `evolver_extra` was the same as `LogInfo.EXTRA_KEY`. The JSON check would still happen in the history hookup, but not at log time.

evolver/logutils.py
amitschang: In order to share this with the history hookup, at that time we do not want to raise an exception, but rather silently drop the data. Users of `LogInfo` ought to always have good data, other users would risk dropping non-jsonable data.

evolver/logutils.py
amitschang: I made update based on your text since it prob read strange (even though I wrote it, not AI 😄 )

imaitland: yeah, no ai from me either.

imaitland (approved):
LGTM.

PR #228, opened 2024-12-10
Add procedure completed actions and undo

imaitland:
- The front end must display the current progress of the procedure. This PR adds a "completed_actions" property.
- The front end must support undoing actions. This PR adds an "undo" endpoint".

### DIFF
--- a/evolver/app/routers/hardware.py
+++ b/evolver/app/routers/hardware.py
@@ -1,6 +1,7 @@
 from typing import Any, Dict, List
 
 from fastapi import APIRouter, Body, Path, Request
+from fastapi.params import Query
 from pydantic import BaseModel
 
 from evolver.app.exceptions import (
@@ -67,11 +68,15 @@ def hardware_commit(hardware_name: str, request: Request):
     hardware_instance.commit()
 
 
-# Start the calibration procedure for the selected hardware and vials
+# Start the calibration procedure for the selected hardware and vials,
+# resume will init the procedure with the CalibrationData from the Calibrator.
+# Where CalibrationData is the state of the procedure that has been persisted to the config file.
+# If resume is False, the procedure state will reset.
 @router.post("/{hardware_name}/calibrator/procedure/start")
 def start_calibration_procedure(
     hardware_name: str,
     request: Request,
+    resume: bool = Query(True),
 ):
     hardware_instance = get_hardware_instance(request, hardware_name)
     calibrator = hardware_instance.calibrator
@@ -80,6 +85,7 @@ def start_calibration_procedure(
 
     calibrator.create_calibration_procedure(
         selected_hardware=hardware_instance,
+        resume=resume,
     )
 
     return calibrator.calibration_procedure.get_state()
@@ -139,6 +145,19 @@ def get_calibrator_state(hardware_name: str, request: Request):
     return calibration_procedure.get_state()
 
 
+# Undo the last action dispatched to the calibration procedure
+@router.get("/{hardware_name}/calibrator/procedure/undo")
+def undo_calibration_procedure_action(hardware_name: str, request: Request):
+    hardware_instance = get_hardware_instance(request, hardware_name)
+    calibrator = hardware_instance.calibrator
+    if not calibrator:
+        raise CalibratorNotFoundError
+    calibration_procedure = calibrator.calibration_procedure
+    return calibration_procedure.undo()
+
+
+# Get the calibrator's CalibrationData, representing the state from the procedure that has been saved
+# This data will appear in the config file even if the procedure is interupted. And will be used as the initial state when the procedure is resumed.
 @router.get("/{hardware_name}/calibrator/data")
 def get_calibration_data(hardware_name: str, request: Request):
     hardware_instance = get_hardware_instance(request, hardware_name)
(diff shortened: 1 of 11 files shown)


evolver/app/routers/hardware.py
```diff
@@ -139,6 +145,19 @@ def get_calibrator_state(hardware_name: str, request: Request):
     return calibration_procedure.get_state()
 
 
+# Undo the last calibration procedure action, reverting the state to the previous state
+@router.post("/{hardware_name}/calibrator/procedure/undo")
+def undo_calibration_procedure_action(hardware_name: str, request: Request):
+    hardware_instance = get_hardware_instance(request, hardware_name)
+    calibrator = hardware_instance.calibrator
+    if not calibrator:
+        raise CalibratorNotFoundError
```
amitschang, line 154: [minor] Would be a good place for walrus!
```py
    if not (calibrator := hardware_instance.calibrator):
        raise CalibratorNotFoundError
```

evolver/app/tests/test_calibration.py
```diff
@@ -72,196 +67,102 @@ def test_temperature_calibration_procedure_actions():
...
+    client.post("/hardware/test/calibrator/procedure/start")
 
-    raw_dispatch_response = client.post(
-        "/hardware/test/calibrator/procedure/dispatch", json={"action_name": "read_vial_0_raw_output", "payload": {}}
-    )
-    assert raw_dispatch_response.status_code == 200
-    assert raw_dispatch_response.json()["0"]["raw"] == [1.23]
+    dispatch_action(client, "test", "measure_vial_0_temperature", {"temperature": 25.0})
+    dispatch_action(client, "test", "read_vial_0_raw_output")
+    dispatch_action(client, "test", "calculate_vial_0_fit")
```
amitschang, line 163: I think this is fine for now, but we might want to make a note to create a mock procedure for the tests which cover action kinds (e.g. ones that need input, ones that don't, etc.), iso using the temperature one - just because this looks more generic

evolver/calibration/interface.py
```diff
@@ -193,5 +193,5 @@ class Config(Calibrator.Config):
         input_transformer: dict[int, ConfigDescriptor | Transformer | None] | None = None
         output_transformer: dict[int, ConfigDescriptor | Transformer | None] | None = None
 
-    def create_calibration_procedure(self, selected_hardware, *args, **kwargs):
+    def create_calibration_procedure(self, selected_hardware, resume, *args, **kwargs):
```
amitschang, line 196: I know this is not added in this PR, but I don't think we even need this method here. It is abstract in the base and we can leave it out to have consistent error (`TypeError` from the ABC) - plus satisfying to remove code!

evolver/calibration/standard/calibrators/temperature.py
```diff
@@ -19,14 +19,26 @@ class TemperatureCalibrator(IndependentVialBasedCalibrator):
 
     class CalibrationData(Transformer.Config):
         measured: Dict[int, Dict[str, List[float]]] = {}  # {vial_index: {"reference": [], "raw": []}}
+        completed_actions: List[str] = []
```
amitschang, line 22: Since `completed_actions` looks like something we want to support at the interface, perhaps this should actually go in `Transformer.Config` (or actually I think better yet: this class should be based off `Calibrator.CalibrationData` and that should have `completed_actions`)

evolver/calibration/procedure.py
```diff
@@ -1,35 +1,43 @@
...
-            Data stored in the CalibrationProcedure state should also be periodically saved to the Calibraor's CalibrationData class.
-            That way CalibrationProcedure state can be saved and reloaded to continue the calibration procedure if interupted.
+            Data stored in the CalibrationProcedure state should also be periodically saved to the Calibrator's CalibrationData class.
+            That way CalibrationProcedure state can be saved and reloaded to continue the calibration procedure if interrupted.
             This can be done by defining a SaveProcedureState action in the procedure and dispatching it.
         """
         super().__init__(*args, **kwargs)
         self.actions = []
-        self.state = {}
+        self.state = CalibrationStateModel(**(state or {})).model_dump()
```
amitschang, line 39: looking at this and the `setdefaults` done below, I wonder if we rather could benefit from `state` being `CalibrationStateModel` iso `Dict` and make the model allow extra fields.

```py
class CalibrationStateModel(BaseModel):
    model_config = ConfigDict(extra='allow')
    completed_actions: List[str] = Field(default_factory=list)
```

This way instantiating will always make the default empty list if not supplied, and user doesn't have to state explicitly deal with that. Also, I think without the extra being "allow"ed, this part might drop the things in `**state` that are not `completed_actions`

amitschang (commented):
overall looks good. Some inline commentsI think can help make it easier to work with if I'm understanding correctly!

evolver/app/routers/hardware.py
imaitland: Nice, will fix.

evolver/calibration/interface.py
imaitland: Ok, cool will fix.

evolver/calibration/standard/calibrators/temperature.py
imaitland: makes sense, will fix.

evolver/calibration/procedure.py
imaitland: will fix.

amitschang (approved):
Thanks for the updates!

PR #243, opened 2025-01-10
vial layout

### LINKED ISSUE #153: Vial physical layout specification in evolver
Right now we have a simple list of the indexes of vials, but this doesn't give enough information about there physical layout unless one assumes the default 16 vial 4x4 grid. 

We should support a configuration that enables both an understanding of the layout of available slots and to specify which are occupied, possibly in a way where the user can assign "names" as needed.

For example:

```
[[0, None, None, None],
 [None, 5, None, None],
 [None, None, 10, None],
 [None, None, None, 15]]
```

or possibly (assuming the numbers don't have to map strictly to top-down-left-right order:

```
[[0, None, None, None],
 [None, 1, None, None],
 [None, None, 2, None],
 [None, None, None, 3]]
```

Assuming we have a 4 x 4 grid but only inhabit one diagonal.

2-d assumption seems appropriate, if needed

amitschang:
This adds configuration and validation for a grid layout specification. I'm not sure at the moment how else this needs to be presented or used in the backend, so it is just the configuration for now, which could be used in frontend for presentation.

resolves #153.

### DIFF
--- a/evolver/device.py
+++ b/evolver/device.py
@@ -1,6 +1,9 @@
 import logging
 import time
 from collections import defaultdict
+from math import prod
+
+from pydantic import Field, ValidationInfo, field_validator
 
 from evolver.base import BaseInterface, ConfigDescriptor
 from evolver.connection.interface import Connection
@@ -20,6 +23,12 @@ class Evolver(BaseInterface):
     class Config(BaseInterface.Config):
         name: str = "Evolver"
         experiment: str = "unspecified"
+        vial_layout: list[int] = Field(
+            default=settings.DEFAULT_VIAL_LAYOUT,
+            description="The layout of the vials in 2 or 3 dimensions. Always left-to-right bottom-top-top order.",
+            min_length=2,
+            max_length=3,
+        )
         vials: list = list(range(settings.DEFAULT_NUMBER_OF_VIALS_PER_BOX))
         hardware: dict[str, ConfigDescriptor | HardwareDriver] = {}
         controllers: list[ConfigDescriptor | Controller] = []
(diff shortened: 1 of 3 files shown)


evolver/device.py
```diff
@@ -20,6 +23,12 @@ class Evolver(BaseInterface):
     class Config(BaseInterface.Config):
         name: str = "Evolver"
         experiment: str = "unspecified"
+        vial_layout: list[int] = Field(
+            default=settings.DEFAULT_VIAL_LAYOUT,
+            description="The layout of the vials in 2 or 3 dimensions. Always left-to-right bottom-top-top order.",
+            min_length=2,
+            max_length=3,
+        )
```
imaitland, line 31: Maybe not for this PR but this cries out for a diagram.

amitschang: you mean like one in docs?

evolver/device.py
```diff
@@ -20,6 +23,12 @@ class Evolver(BaseInterface):
     class Config(BaseInterface.Config):
         name: str = "Evolver"
         experiment: str = "unspecified"
+        vial_layout: list[int] = Field(
+            default=settings.DEFAULT_VIAL_LAYOUT,
+            description="The layout of the vials in 2 or 3 dimensions. Always left-to-right bottom-top-top order.",
```
imaitland, line 28: Another note, does this incorporate the grids of vials within grids of vials that were shown by Nate?

evolver/device.py
imaitland: yeah, i notice the issue has some more visual description of the idea.

evolver/device.py
amitschang: if you mean the one where they piggyback a group of smaller vials together on one larger slot, then yes I don't see why not. The addressing would be handled at the hardware driver level, and this would map the layout of how those are situated.

imaitland: nice

imaitland (approved):
LGTM

PR #244, opened 2025-01-23
experiment layer wip

amitschang:


evolver/app/tests/test_app.py
```diff
@@ -256,6 +256,50 @@ def read(self):
...
+            }
+        )
+        response = app_client.get("/experiment/test/logs")
+        assert response.status_code == 200
+        assert response.json() == {"data": {}}
+        app.state.evolver.loop_once()
+        response = app_client.get("/experiment/test/logs")
+        assert len(response.json()["data"]["NoOpController1"]) == 1
+        assert len(response.json()["data"]["NoOpController2"]) == 1
+
```
imaitland, line 302: A test for /experiment/test too?

evolver/app/routers/experiment.py
```diff
@@ -0,0 +1,26 @@
...
+    evolver = request.app.state.evolver
+    controllers = evolver.experiments[experiment_name].controllers
+    names = [c.name for c in controllers if c.name]
+    return evolver.history.get(names=names, kinds=["log", "event"])
+
+
+@router.get("/{experiment_name}")
+def get_experiment_overview(request: Request, experiment_name: str):
+    return {
+        "spec": request.app.state.evolver.experiments[experiment_name],
```
imaitland, line 24: is "spec" a config?

evolver/device.py
```diff
@@ -19,10 +19,23 @@
...
+    controllers: list[ConfigDescriptor | Controller] = []
+
+    # this seemed to have been required for the tests of config symmetry.
+    # Without it there is pydantic error about unkown type (the underlying
+    # Controller class) - so seemed to be kind of bypassing the evolver
+    # BaseModel hookups?
+    @field_serializer("controllers")
+    def serialize_controllers(self, data):
+        return [ConfigDescriptor.model_validate(c) for c in data]
+
```
imaitland, line 33: probably worth getting @jamienoss feedback here? to at least have more detail / as much detail as possible in the comment, especially if there's something that's not obeying the underlying config model?

evolver/device.py
```diff
@@ -53,6 +66,14 @@ def validate_vials(cls, value: list[int], info: ValidationInfo):
     def __init__(self, *args, **kwargs):
         self.last_read = defaultdict(lambda: int(-1))
         super().__init__(*args, evolver=self, **kwargs)
+        # We have to turn experiment controllers passed in as configdescriptors
+        # to objects while passing in self so controllers can consume hardware
+        # and read from history, etc.
+        for experiment in self.experiments.values():
+            for i in range(len(experiment.controllers)):
+                elem = experiment.controllers[i]
+                if isinstance(elem, ConfigDescriptor):
```
imaitland, line 75: Is this reflected in the type definition `experiments: dict[str, Experiment] = {}` like is `Experiment` typed somewhere as a ConfigDescriptor? Line of reasoning, if at somepoint `elem` here stops being an instance of ConfigDescriptor this condition will silently fail (i maybe wrong in the silent part) unless some typing also accompanies the condition....

evolver/device.py
```diff
@@ -76,6 +97,10 @@ def sensors(self):
     def effectors(self):
         return {k: v for k, v in self.hardware.items() if isinstance(v, EffectorDriver)}
 
+    @property
+    def active_controllers(self):
+        return [c for exp in self.experiments.values() for c in exp.controllers if exp.enabled]
```
imaitland, line 102: bit of a mismatch `active` determined by `enabled` property, `def enabled_controllers(self):` would be more consistent. Would also be nice to have a doc string.

evolver/history/interface.py
```diff
@@ -51,6 +52,9 @@ def get(
         Args:
             name: The name of hardware component to retrieve data for. If None,
                 return all data.
+            names: A list of names of hardware components to retrieve data for.
+                Should be Mutually exclusive with the name parameter, where the
+                name parameter would take precedence if supplied.
```
imaitland, line 57: hmmm... was name kept for backwards compatibility? why wouldn't someone just pass "names" with just one name in it if they were getting a single item?

If I pass both name and names parameters, the implicit behavior is going to be confusing... when it comes to the response.

evolver/app/routers/experiment.py
```diff
@@ -0,0 +1,26 @@
+from fastapi import APIRouter, Request
+
+from evolver.device import Experiment
+
+router = APIRouter(prefix="/experiment", tags=["experiment"])
+
+
+@router.get("/")
```
imaitland, line 8: i wonder if this api layout makes assumptions about querying for one experiment at a time (though the underling framework does support getting data for multiple experiments at once?

Could be nothing, especially if there's no use case for asking for logs for multiple experiment_names at a time.

imaitland (commented):
Looking forward to building out the UI for this. Added some comments.

evolver/app/routers/experiment.py
amitschang: Yeah, I think it ought to get serialized to a configdescriptor, that is the intention. Do you think there is a better name, or should have a model defined in code?

evolver/device.py
amitschang: this is from def of experiment.controllers:

```py
class Experiment(BaseInterface.Config):
    enabled: bool = True
    controllers: list[ConfigDescriptor | Controller] = []
```

It is expected to possibly be a configdescriptor. I don't think there is a case for silent failure, if a proper model instance is passed, pydantic will do the validation, otherwise the object will either have to implement the `Controller` API, or there will be an exception because it doesn't, which is what we expect.

evolver/history/interface.py
amitschang: Yeah, for that and to limit the blast radius of the PR, but I agree: we can rather just remove name and I'll modify the tests in accordance, there shouldn't really be a need to support both and the work should be done sometime

evolver/app/routers/experiment.py
amitschang: sorry, not a configdescriptor, it will get serialized to a `Experiment` representation, like:
```
{
  "controllers": [ ConfigDescriptor, ... ],
  "enabled": ...
  "name": ...
}
```

imaitland: yeah i think because we call these "configs" generally and not "specs", I think ```return { "config": request.app...}``` makes more sense to me.

imaitland (approved):
LGTM

PR #246, opened 2025-01-23
All json serialization does so to ConfigDescriptor (WIP)

### LINKED ISSUE #213: Bug in nested rebuilding of Config as ConfigDescriptor

jamienoss:
Resolves #213

### DIFF
--- a/evolver/base.py
+++ b/evolver/base.py
@@ -92,18 +92,28 @@ class _BaseModel(pydantic.BaseModel):
 
 class _BaseConfig(_BaseModel):
     @classmethod
-    def get_classinfo(cls) -> str:
+    def get_classinfo(cls, strict=True) -> str:
         """The fully qualified classname for cls's container class (if one exists).
         E.g., ``Evolver.Config.get_classinfo()`` returns ``"evolver.device.Evolver"``.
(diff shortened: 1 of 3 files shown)


evolver/base.py
```diff
@@ -92,19 +92,29 @@ def model_validate(cls, obj, *, strict=None, from_attributes=None, context=None)
...
             cls = container_class
         return f"{cls.__module__}.{cls.__qualname__}"
 
+    @pydantic.model_serializer(mode="wrap", when_used='json')
+    def to_descriptor(self, handler) -> dict:
+        serialized_by_super = handler(self)
+        class_info = self.get_classinfo()
+        if class_info is None or _is_descriptor_dict(serialized_by_super):
+            return serialized_by_super
+        return dict(classinfo=class_info, config=serialized_by_super)
```
jamienoss, line 116: @amitschang This is the serializer. Most of the other changes in this PR might be temp and I just committed and pushed what I had.

jamienoss:
@amitschang feel free to close this or keep it.

PR #253, opened 2025-02-06
resume procedure endpoint

imaitland:
feature - when a calibration procedure state has been saved (persisted to fs), it can be 'resumed' by passing the "resume" flag to the /start endpoint.


evolver/app/routers/hardware.py
```diff
@@ -135,7 +140,7 @@ def dispatch_calibrator_action(request: Request, hardware_name: str = Path(...),
 
     payload = action.get("payload", {})
 
-    return {**calibration_procedure.dispatch(action_to_dispatch, payload), "started": True}
```
imaitland, line 138: Refactored "started" atrribute to be common state for all procedures. It's needed as the procedure controls (UI) are distinct for an uninitialized (not started) and initialized (started) procedure.

evolver/app/tests/test_calibration.py
```diff
@@ -44,7 +44,7 @@ class TestCalibration:
     def test_get_calibration_status(self):
         _, client = setup_evolver_with_calibrator(NoOpCalibrator)
 
-        response = client.post("/hardware/test/calibrator/procedure/start")
+        response = client.post("/hardware/test/calibrator/procedure/start", params={"resume": False})
```
imaitland, line 47: Now the user must be explicit about their intention to attempt to resume a procedure or start a new / fresh one.

evolver/app/tests/test_calibration.py
```diff
@@ -235,27 +265,31 @@ def test_get_calibration_data(tmp_path):
     assert calibration_data_response.status_code == 200
     calibration_data = calibration_data_response.json()
 
-    assert calibration_data["measured"] == {
+    assert calibration_data["procedure_state"] == {
```
imaitland, line 268: renamed "measured" -> "procedure_state" for the TemperatureCalibrator,  since this is what it is.

evolver/app/tests/test_calibration.py
```diff
@@ -107,15 +107,15 @@ def test_dispatch_temperature_calibration_raw_value_action():
...
+    client.post("/hardware/test/calibrator/procedure/start", params={"resume": False})
 
     raw_dispatch_response = dispatch_action(client, "test", "read_vial_0_raw_output")
 
     assert raw_dispatch_response.status_code == 200
     assert raw_dispatch_response.json() == {
         "0": {"raw": [1.23], "reference": []},
         "completed_actions": ["read_vial_0_raw_output"],
-        "history": [{"completed_actions": [], "history": []}],
+        "history": [{"completed_actions": [], "history": [], "started": True}],
```
imaitland, line 118: only downside to adding "started" to common procedure state is it appears in history.

evolver/calibration/interface.py
```diff
@@ -119,9 +119,13 @@ class Config(Transformer.Config):
         output_transformer: ConfigDescriptor | Transformer | None = None
         calibration_file: str | None = None
 
-    class CalibrationData(Transformer.Config, CalibrationStateModel):
-        """Stores calibration data, including the completed_actions and the "measured" data the actions have collected
-        from the CalibrationProcedure."""
+    class CalibrationData(Transformer.Config):
+        """Stores calibration data, including the procedure_state from the calibration_procedure."""
+
+        procedure_state: CalibrationStateModel = Field(
```
imaitland, line 125: CalibrationStateModel is now common model shared across the Procedure, Actions and CalibrationData

evolver/calibration/procedure.py
```diff
@@ -1,30 +1,9 @@
 from abc import ABC
-from copy import deepcopy
-from typing import Any, Dict, List
-
-from pydantic import BaseModel, Field
+from typing import Any, Dict
 
 from evolver.base import BaseInterface
 from evolver.calibration.action import CalibrationAction
-
```
imaitland, line 9: moved this into the state.py file to avoid circular import errors.

evolver/calibration/procedure.py
```diff
@@ -75,29 +55,31 @@ def undo(self):
         """
         Undo the last action that was dispatched in the calibration procedure.
         """
-        if len(self.state["history"]) > 0:
-            self.state = self.state["history"].pop()
+        if len(self.state.history) > 0:
```
imaitland, line 58: since state remains a pydantic model now across the system, we can get typed property access.

evolver/app/tests/test_calibration.py
imaitland: that's not true, ignore.

evolver/calibration/procedure.py
```diff
@@ -54,7 +33,8 @@ def __init__(self, hardware, state=None, *args, **kwargs):
         """
         super().__init__(*args, **kwargs)
         self.actions = []
-        self.state = CalibrationStateModel(**(state or {})).model_dump()
+        self.state = CalibrationStateModel(**(state or {}))
```
amitschang, line 36: It's not wrong, but since we are working with an object, might it be cleaner to
```py
CalibrationStateModel.model_validate(state or {})
```

And I didn't really catch earlier, but since the init takes state as a parameter we could default `state={}` there, and can pass just `state` to the above - unless of course we specifically need to support None being passed in.

evolver/calibration/standard/calibrators/temperature.py
```diff
@@ -17,7 +16,7 @@ class TemperatureCalibrator(IndependentVialBasedCalibrator):
     """
 
     class CalibrationData(Calibrator.CalibrationData):
-        measured: Dict[int, Dict[str, List[float]]] = {}  # {vial_index: {"reference": [], "raw": []}}
+        procedure_state: CalibrationStateModel = {}
```
amitschang, line 19: in this case we actually don't even need to overload anymore, so this whole class can (and should) be removed.

evolver/calibration/procedure.py
imaitland: yep will fix

evolver/calibration/standard/calibrators/temperature.py
imaitland: yep will fix

evolver/app/tests/test_calibration.py
```diff
@@ -212,6 +275,22 @@ def test_dispatch_temperature_calibration_calculate_fit_action():
     assert output_transformer_response.json()["0"]["parameters"] == [0.6149999999999997, 0.024599999999999997]
 
 
+def get_stable_state_subset(state):
```
imaitland, line 278: Having to do this reveals an area for improvement in the history data structure. Since history is part of the state persisted to the history, the entire history is actually already available in the history prop of the last added item to the history stack...

Soooo...
To fix, update the history stack data-structure update methods (push and pop) to only need the last state in the stack, and update the undo method, to essentially walk the history attr of the single state that's now in the history stack.

See issue: https://github.com/ssec-jhu/evolver-ng/issues/256

evolver/calibration/interface.py
```diff
@@ -144,7 +162,10 @@ def __init__(self, *args, calibration_file=None, **kwargs):
             self.calibration_file = calibration_file
             self.load_calibration_file(calibration_file)
         else:
-            self.calibration_data = self.CalibrationData()
+            if self.calibration_file:
+                self.calibration_data = CalibrationStateModel.load(file_path=self.calibration_file)
+            else:
+                self.calibration_data = CalibrationStateModel()
```
amitschang, line 168: I think this is not quite right. Since this is init and we are passed the calibration_file to input, we shouldn't have the `self.calibration_file` unless set here, so the above condition (`if calibration_file`) should be sufficient?  Meaning here is just to create a default calibration state

evolver/calibration/standard/actions/temperature.py
```diff
@@ -1,26 +1,25 @@
...
 
-    def execute(self, state: Dict, payload: Optional[FormModel] = None):
-        state.setdefault(self.vial_idx, {"reference": [], "raw": []})
-        state[self.vial_idx]["reference"].append(payload.temperature)
+    def execute(self, state: CalibrationStateModel, payload: Optional[FormModel] = None):
+        if not state.measured:
+            state.measured = {}
+        if self.vial_idx not in state.measured:
+            state.measured[self.vial_idx] = {"reference": [], "raw": []}
+        state.measured[self.vial_idx]["reference"].append(payload.temperature)
```
amitschang, line 22: totally fine, but in case you like: `collections.defaultdict` and/or `setdefault` is useful here:

```py
state.measured = state.measured or defaultdict(lambda: {"referece": [], "raw": []})
state.measured[self.vial_idx]["reference"].append(payload.temperature)
```

where all elements are same kind, no need to check and initialize yourself.

amitschang (approved):
Thanks! I think it can go, but a couple comments for your perusal

evolver/calibration/interface.py
imaitland: Yeah makes sense. Fixed.

amitschang (approved):

PR #255, opened 2025-02-11
dev endpoint

imaitland:
- Move the `/experiment` tests into their own file to match how each router has their own folder for tests
- add a `/dev` endpoint that exposes a http endpoint to manually tick the evolver loop - useful for generating log entries against demo / noop controllers - which the UI needs to build out demo pages.

I envision a "demo" evolver running in the cloud somewhere, with NoOpControllers and mock data, that the UI can connect to by default anytime, in a 'read-only' way to demonstrate the system or indeed test it end-to-end...

Plays on top of the commits in the /resume procedure PR. Will rebase this PR once that's squashed and merged.


imaitland:
not needed.

PR #264, opened 2025-02-26
falling back to None for uncalibrated values, iso raw value

### LINKED ISSUE #263: calibration fallback should result in null, not untransformed raw value
from Ezira: 

>  i think it makes sense for raw sensor states to still be stored in the event of a failed calibration rather than not saving the raw completely (maybe show a null in the calibrated data rather than just copying the raw to distinguish between situations where calibration is not provided) (edited)

amitschang:
Implemented as a configurable fallback value which defaults to `None`

resolves #263


imaitland (approved):
LGTM!

PR #267, opened 2025-02-26
separate ongoing procedure file from applied calibration file

### LINKED ISSUE #261: Sort out calibration file lifecycle for procedure and application
At present, the calibration procedure uses the calibrator's `calibration_file` attribute, but this is the same that specifies the file to use for transforming, which upon loading expects a file to exist and have contents used in the fit/transformation - leading to a chicken-egg scenario. 

A suggestion:

As the calibration procedure is that which creates a calibration file to be applied, have the procedure take a filename at start, run the procedure, then prompt the user to "apply" this calibration the effect of which would be to set `calibration_file` on calibrator to the one supplied to the procedure at start.

a basic visualization of this:

```
create_calibration_procedure(filename) -> run_all_seteps -> output_calibration_file -> apply_calibration -> calibrator.set_calibration_file_in_config
```

amitschang:
Keep ongoing/partial procedure state to a separate file to avoid startup conflict with calibration file loading (if specified, the calibration file expects a complete file to load at startup, but the procedure necessarily expects and empty one at start, or a partial one at resume).

Resolves #261

### DIFF
--- a/evolver/app/tests/test_calibration.py
+++ b/evolver/app/tests/test_calibration.py
@@ -11,7 +11,7 @@ from evolver.hardware.demo import NoOpSensorDriver
 from evolver.tests.conftest import tmp_calibration_dir  # noqa: F401
 
 
-def setup_evolver_with_calibrator(calibrator_class, hardware_name="test", vials=[0, 1, 2], calibration_file=None):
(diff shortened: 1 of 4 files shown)


imaitland (approved):
LGTM!

Issue #234, opened 2024-12-17
Can't pass calibration config-descriptor file via evolver.yml config


amitschang:
I'm not sure if this is actually supposed to work even, but I tried to specify the fitted calibration file to an od sensor via the following config:

```yaml
  od:
    classinfo: evolver.hardware.standard.od_sensor.ODSensor
    config:
      addr: od_90
      integrations: 500
      calibrator: /calibration_files/default_od90_calibration_config.yml
      name: ODSensor

```

but I get the following error:

```
Dec 17 20:51:30 raspberrypi python[856062]: pydantic_core._pydantic_core.ValidationError: 2 validation errors for Config
Dec 17 20:51:30 raspberrypi python[856062]: calibrator.ConfigDescriptor
Dec 17 20:51:30 raspberrypi python[856062]:   Input should be a valid dictionary or object to extract fields from [type=model_attributes_type, input_value='/calibration_file
s/defau..._calibration_config.yml', input_type=str]
Dec 17 20:51:30 raspberrypi python[856062]:     For further information visit https://errors.pydantic.dev/2.7/v/model_attributes_type
Dec 17 20:51:30 raspberrypi python[856062]: calibrator.Config
Dec 17 20:51:30 raspberrypi python[856062]:   Input should be a valid dictionary or object to extract fields from [type=model_attributes_type, input_value='/calibration_file
s/defau..._calibration_config.yml', input_type=str]
Dec 17 20:51:30 raspberrypi python[856062]:     For further information visit https://errors.pydantic.dev/2.7/v/model_attributes_type
Dec 17 20:51:30 raspberrypi python[856062]: ERROR:    Application startup failed. Exiting.
```

amitschang:
#268 makes this no longer necessary I think

Issue #193, opened 2024-10-16
Make Transformer parameters optional and fallback to Calibrator.measured data to then gen transformer params


jamienoss:
We want Calibrators and Transformers to be trivially configurable and thus swappable. To do this, we need to make ``Transfomer.Config.parameters`` optional such that only some config fields are populated, e.g., polyfit degree, but not the actual polynomial coefficients.

Then, a calibrator would need ``measured_data`` (relates to #165) to call ``Transformer.fit`` to fit the measured data and generate the desired transformer parameters.

If both transformer optional params and calibrator optional measured data aren't present, obviously raise.

jamienoss:
@imaitland FYI

amitschang:
We have the tools now for measured data to be stored and to do refit on initialization and during procedure using classes provided. See #268 and references therein

Issue #179, opened 2024-10-07
Unbake calibration_file attr from calibration config.


jamienoss:
Relates to #174 

See https://github.com/ssec-jhu/evolver-ng/pull/174#discussion_r1790237469

All configs can be loaded from file using ``model_validate``, the calibration files are no exception.

See https://github.com/ssec-jhu/evolver-ng/blob/main/evolver/base.py#L52-L53
```python
class _BaseConfig(pydantic.BaseModel):
    model_config = pydantic.ConfigDict(extra="ignore", from_attributes=True)

    @classmethod
    def model_validate(cls, obj, *, strict=None, from_attributes=None, context=None):
        """Effectively the same as pydantic.BaseModel.model_validate() except that it automatically handles json, and
        conversion from instances of ``BaseConfig`` and ``BaseInterface`
        """
        if obj is None:
            return cls()
        elif isinstance(obj, os.PathLike):
            return cls.load(file_path=obj)
        ...
```

It's quite possible that there are a few code paths that need patching or re-directing to ensure that this works from all nominal invocations, # can contain that if needed.

jamienoss:
@amitschang FYI

jamienoss:
Relates to #165 as the ``calibration_file`` _is_ ``CalibrationData``.

jamienoss:
One negative aspect of doing this would be the calibration data significantly bloating the overall ``Evolver.Config``. We could special case the model conversion of this so that it always gets serialized to a filepath, that also gets saved to disk (on the PI).
A better solution, to avoid the special casing, could be to add such conversion for all and add a setting, e.g., ``MAX_CONFIG_SECTION_LINE_LIMIT``, that when a nested section exceeds this number of lines/entries it is written to disk and "serialized" in the resultant config json/yaml as a filepath. Obviously, this isn't true serialization and would require some mechanism for copying said sub-files back and forth to the server & client.

amitschang:
We are going to keep calibration_file attribute and have a mechanism for creating via procedure and for loading and specifying the file. Fully spec'ed calibration can still be inlined and null calibration file is legal.

Issue #165, opened 2024-09-04
Store measured data from calibration in Calibrator.Config


jamienoss:
Punt from #157 

See https://github.com/ssec-jhu/evolver-ng/pull/157#issuecomment-2327255025

```python
        measured_data: dict | None = Field(
            None,
            description="The data measured by a calibration procedure run. This can be specified instead of specifying "
            "some transformer config params, e.g., polynomial coefficients. This data will be fitted upon "
            "first use.",
        )
```

imaitland:
Worked this into the calibration procedure pr with Jamie's feedback.

amitschang:
we store the measured data in calibration state model object, typically written via procedure. See also #253 and #268

PR #268, opened 2025-02-27
Store fitted calibrator in state model, path to load, and default transformers

amitschang:
This change adds:
* A place to store fitted transformers within the calibration procedure state model, enabling to use the prefit transformers without a refit (in cases where no measured data is stored, or when fitting must be done outside the system).
* A path to detect when to use fitted transformers and automatically apply
* Parameters for default transformers for the multi-vial case. This can be used in the common case where all vials use the same method (each will still be fit with their own parameters)
* Code for at-initialization refit of the temperature calibrator.

I think this resolves or otherwise closes following issues:
* #234 since we can pass via `calibration_file` property now
* #179 since we are committed to working with the files for procedure and loading
* #165 I think this is actually resolved already, but this closes the loop in how to apply from saved procedure data
* #193 we have the machinery to refit to a transformer config spec'ed without fit params

### DIFF
--- a/evolver/calibration/demo.py
+++ b/evolver/calibration/demo.py
@@ -13,6 +13,10 @@ class NoOpTransformer(Transformer):
     def convert_to(self, data, *args, **kwargs):
         return data
 
+    def refit(self, *args, **kwargs):
+        self._refit_args = args
+        self._refit_kwargs = kwargs
+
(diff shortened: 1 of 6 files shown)


imaitland (approved):
LGTM!

Issue #182, opened 2024-10-09
UI feedback/wishlist


amitschang:
From slack from Nate, made into checkboxes in case they are todos or turn into issues, otherwise strike/comment as needed.

* State Page
  * [ ] Toggle between Raw / Calibrated value display
  * [ ] Marking which controllers are active on which vials
  * [ ] Coloring the border of the vial and showing a key on sidebar
* Hardware page
  * [ ] Is this where current experiments will be displayed? Or is it only showing history server output?
* Experiment page?
  * [ ] Not sure if separate from hardware page
  * [ ] Organized graphs by experiment
  * [ ] Like you guys said - logged events on graphs
  * [ ] Easy zooming / scaling of graphs
  * [ ] Display of current parameters for this experiment such as flow rate for chemostat and OD upper and lower bounds for turbidostat
  * [ ] Easy export of data
* Controller page?
  * [ ] Not 100% sure what will eventually go on controller page
  * [ ] Config editor where we don't have to edit each value of config individually

nb-e:
Experiment page wants, but not essentials
- [ ] Graphs toggle between line and scatter plots
- [ ] Can copy graph as a png or download data from this view

PR #269, opened 2025-02-28
Iainland/controller config attr endpoint

imaitland:
Adds controller config to the experiments endpoint, needed for the UI to meet req: 
>  Display of current parameters for this experiment such as flow rate for chemostat and OD upper and lower bounds for turbidostat

from: https://github.com/ssec-jhu/evolver-ng/issues/182

### DIFF
--- a/evolver/app/routers/experiment.py
+++ b/evolver/app/routers/experiment.py
@@ -23,18 +23,7 @@ def get_experiment_overview(request: Request, experiment_name: str):
     evolver = request.app.state.evolver
     experiment = evolver.experiments[experiment_name]
 
-    # Extract controller configs and current states
-    controller_data = []
-    for controller in experiment.controllers:
-        controller_info = {
-            "name": controller.name,
-            "type": controller.__class__.__name__,
-            "config": controller.config_model.model_dump(),
-        }
-        controller_data.append(controller_info)
-
     return {
         "config": experiment,
         "logs": get_experiment_logs(request, experiment_name),
-        "controllers": controller_data,
     }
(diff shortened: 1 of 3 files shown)


evolver/app/routers/experiment.py
```diff
@@ -20,7 +20,21 @@ def get_experiment_logs(request: Request, experiment_name: str):
...
+
+    # Extract controller configs and current states
+    controller_data = []
+    for controller in experiment.controllers:
+        controller_info = {
+            "name": controller.name,
+            "type": controller.__class__.__name__,
+            "config": controller.config_model.model_dump(),
+        }
+        controller_data.append(controller_info)
```
amitschang, line 34: I think the config field already does all that is needed, unless I am missing something here. If you remove this block the tests still succeed. Is there some bit of info that is missing?

evolver/app/routers/experiment.py
```diff
@@ -20,7 +20,21 @@ def get_experiment_logs(request: Request, experiment_name: str):
...
+
     return {
-        "config": request.app.state.evolver.experiments[experiment_name],
+        "config": experiment,
         "logs": get_experiment_logs(request, experiment_name),
+        "controllers": controller_data,
```
amitschang, line 39: likewise, I think this information is available in config above, it is structured as:

```
config:
  controllers: 
    - ...
```

evolver/app/routers/experiment.py
imaitland: ah duh, yeah. What happened was I checked the response on the NoOpController, saw it was missing any config attributes like "stir_rate", added some, wrote this endpoint to access them at "controllers", and then had ChatGpt write a test which it did against the actual nested Controller config...

thanks... lesson - check AI code...

amitschang (approved):
Looks good, thanks!

PR #274, opened 2025-03-17
add name to healthz endpoint

imaitland:
adds the evolver name to the healthz response. This is so the frontend can use a human-readable name for the device as it exists in the frontend database.


amitschang (approved):
lgtm

PR #278, opened 2025-03-19
fallback in transform on exception

### LINKED ISSUE #266: Transformer fallback to fallback value when exception in transformation
Seems the desire is if there is an exception attempting the transformation (which could be in custom code), that the resultant value returns null. At present the exception may be causing the read to fail before returning any outputs.

amitschang:
resolves #266


imaitland (approved):
LGTM

PR #280, opened 2025-03-25
update to hardware effector set method

amitschang:
I noticed this for a while but was reminded again when working with the pump driver. In some cases we might be working with the model when setting values to hardware, but in other case removing the need to explicitly construct the model is more natural. This enables both:

```py
model = Driver.Input(vial=0, value=1)
hardware.set(model)
```

or 
```py
hardware.set(vial=0, value=1)
```

With this change, these are equivalent (assuming hardware is instance of `Driver`) and both go through model validation.

Comes with some updated doc strings as well


imaitland (approved):
LGTM

amitschang:
Thanks!

PR #281, opened 2025-03-25
Pump calibration procedure

amitschang:
This PR adds a procedure to the generic pump calibrator, while resolving some minor issues related to the pump necessary for calibration (e.g. setting `time` directly) and adding some conveniences around creating the procedure.

Specifically:
* Lift-and-shift the generic pump calibrator to a module under `standard/calibrators`
* Add the pump calibration procedure, with custom actions
* Update pump input model to take `time` (when `volume` not given)
* Add spec for active pumps on driver (slots refers to total address space, active refers to ones used)
* ~Return `self` in procedure `add_action` enabling operation chaining for readability~
* ~Add `add_action_list` method to procedure, enabling adding list in single chained call~
* Add method for getting an action by name (`get_action(self, name)`)

### DIFF
--- a/evolver/calibration/procedure.py
+++ b/evolver/calibration/procedure.py
@@ -44,9 +44,21 @@ class CalibrationProcedure(BaseInterface, ABC):
                 f"If you want to repeat an action, any action can be dispatched multiple times using the HTTP api."
             )
         self.actions.append(action)
+        return self
+
+    def add_action_list(self, actions: list[CalibrationAction]):
+        for action in actions:
+            self.add_action(action)
+        return self
 
     def get_actions(self):
         return self.actions
+
+    def get_action(self, name):
+        actions = [action for action in self.actions if action.name == name]
+        if len(actions) == 0:
+            raise ValueError(f"Action with name '{name}' not found.")
+        return actions[0]
 
     def get_state(self, *args, **kwargs):
         return self.state
@@ -101,7 +113,9 @@ class CalibrationProcedure(BaseInterface, ABC):
 
         return self.state
 
-    def dispatch(self, action: CalibrationAction, payload: Dict[str, Any]):
+    def dispatch(self, action: CalibrationAction | str, payload: Dict[str, Any]):
+        if isinstance(action, str):
+            action = self.get_action(action)
         if payload is not None and action.FormModel.model_fields != {}:
             payload = action.FormModel(**payload)
         previous_state = self.state.model_dump()
--- a/evolver/calibration/procedure.py
+++ b/evolver/calibration/procedure.py
@@ -107,7 +107,7 @@ class CalibrationProcedure(BaseInterface, ABC):
 
         return self.state
 
-    def dispatch(self, action: CalibrationAction | str, payload: Dict[str, Any]):
+    def dispatch(self, action: CalibrationAction, payload: Dict[str, Any]):
         if payload is not None and action.FormModel.model_fields != {}:
             payload = action.FormModel(**payload)
         previous_state = self.state.model_dump()
--- a/evolver/calibration/procedure.py
+++ b/evolver/calibration/procedure.py
@@ -44,12 +44,6 @@ class CalibrationProcedure(BaseInterface, ABC):
                 f"If you want to repeat an action, any action can be dispatched multiple times using the HTTP api."
             )
         self.actions.append(action)
-        return self
-
-    def add_action_list(self, actions: list[CalibrationAction]):
-        for action in actions:
-            self.add_action(action)
-        return self
 
     def get_actions(self):
         return self.actions
@@ -114,8 +108,6 @@ class CalibrationProcedure(BaseInterface, ABC):
         return self.state
 
     def dispatch(self, action: CalibrationAction | str, payload: Dict[str, Any]):
-        if isinstance(action, str):
-            action = self.get_action(action)
         if payload is not None and action.FormModel.model_fields != {}:
             payload = action.FormModel(**payload)
         previous_state = self.state.model_dump()
--- a/evolver/calibration/standard/calibrators/pump.py
+++ b/evolver/calibration/standard/calibrators/pump.py
@@ -65,51 +65,47 @@ class GenericPumpCalibrator(IndependentVialBasedCalibrator):
 
     def create_calibration_procedure(self, selected_hardware, resume, *args, **kwargs):
         procedure_state = CalibrationStateModel.load(self.procedure_file) if resume else None
-
-        self.calibration_procedure = (
-            CalibrationProcedure(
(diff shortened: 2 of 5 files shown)


evolver/hardware/standard/pump.py
```diff
@@ -1,39 +1,15 @@
...
 from evolver.base import BaseConfig, ConfigDescriptor
-from evolver.calibration.interface import Calibrator, IndependentVialBasedCalibrator, Transformer
+from evolver.calibration.interface import Calibrator
+from evolver.calibration.standard.calibrators.pump import GenericPumpCalibrator
 from evolver.hardware.interface import EffectorDriver
 from evolver.hardware.standard.base import SerialDeviceConfigBase
 from evolver.serial import SerialData
 
 
-class RateTransformer(Transformer):
```
amitschang, line 12: this chunk removal is code moved to standard/calibrators/pump.py

evolver/tests/test_logging.py
```diff
@@ -20,7 +20,11 @@ def test_log_capture_handler():
     assert hist.kind == "log"
 
 
-def test_log_evolver_hookup():
+def test_log_evolver_hookup(caplog):
```
amitschang, line 23: pytest was capturing the log and not passing it on, but we specifically want it for this test.

evolver/calibration/procedure.py
```diff
@@ -101,7 +113,9 @@ def apply(self):
 
         return self.state
 
-    def dispatch(self, action: CalibrationAction, payload: Dict[str, Any]):
+    def dispatch(self, action: CalibrationAction | str, payload: Dict[str, Any]):
+        if isinstance(action, str):
+            action = self.get_action(action)
```
imaitland, line 118: this violates dispatch being a pure function... as it adds "get_action" as a side effect.

evolver/calibration/procedure.py
```diff
@@ -44,10 +44,22 @@ def add_action(self, action: CalibrationAction):
                 f"If you want to repeat an action, any action can be dispatched multiple times using the HTTP api."
             )
         self.actions.append(action)
+        return self
+
+    def add_action_list(self, actions: list[CalibrationAction]):
```
imaitland, line 49: i'm a little sceptical of this, as it expands the DSL feature set to include looping. I think if you're writing a procedure, python's looping syntax should be sufficiently familiar. I'm wary of creating a new looping syntax.

evolver/calibration/standard/calibrators/pump.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def convert_from(self, volume):  # convert to time from volume
+        return volume / self.rate
+
+
+class PumpAction(CalibrationAction):
+    """Run the pump for the configured time.
+
+    The action takes as input a flag for running in fast mode, the times of
+    which are configured in the constructor fast/slow args (in seconds).
```
imaitland, line 23: Actions must be pure. The correct place for fast/slow speed values is in the procedure state.

evolver/calibration/standard/calibrators/pump.py
```diff
@@ -0,0 +1,115 @@
...
+                    description="Run pumps",
+                    hardware=selected_hardware,
+                )
+            )
+            .add_action(
+                DisplayInstructionAction(
+                    description="Wait for pumps to finish", name="wait_pumps", hardware=selected_hardware
+                )
+            )
+            .add_action_list(
```
imaitland, line 104: i appreciate the commitment to an un-interuppted dot chain but I think leaving looping to python is the right move here.

evolver/hardware/standard/pump.py
```diff
@@ -50,14 +26,23 @@ class GenericPump(EffectorDriver):
     class Config(SerialDeviceConfigBase, EffectorDriver.Config):
         ipp_pumps: bool | list[int] = Field(False, description="False (no IPP), True (all IPP), or list of IPP ids")
         calibrator: ConfigDescriptor | Calibrator = ConfigDescriptor(classinfo=GenericPumpCalibrator)
+        active_pumps: list[int] | None = Field(None, description="List of active pump IDs, or None for all")
```
imaitland, line 29: bit of a red flag there defaulting "none" to all will be very confusing imo... 

Did you look at an intermediate "class" that could "enhance" it's children with features specific to some of the assumptions being made - similar to IndependentVialBasedCalibrator. In that case it's being upfront that independent vials are being assumed, here it seems an analog would be "independent pumps"?

evolver/hardware/standard/pump.py
```diff
@@ -67,7 +52,11 @@ def __init__(self, *args, **kwargs):
     def serial(self):
         return self.serial_conn or self.evolver.serial
 
-    def set(self, input):
+    def set(self, *args, **kwargs):
+        # We index here by pump_id, which is a concept outside of vail - we
+        # don't want confuse things by overloading the concept, so be explicit
+        # here
+        input = self._get_input_from_args(*args, **kwargs)
```
imaitland, line 59: see above comment

evolver/calibration/procedure.py
amitschang: Actually, I think 
```py
[Action(...) for i in ...]
```
is typically considered more "pythonic" than
```py
for i in ...:
   Action(...)
```
In either case, the looping is still done in python, but we have enabled the list-comprehension to be used and obviate the need for passing around intermediate reference to the object.

evolver/calibration/standard/calibrators/pump.py
imaitland: yes, fast/slow timings must be added to the initial state of the procedure if actions use those values.

It's very important that actions be pure. I know it may seem like a useless distinction, especially in an OOP context like this, and indeed some actions violate this rule but violations must be avoided.

Functional purity / idempotency is a characteristic that is integral to the underlying philosophy / pattern that the calibration procedures are modeled on: https://elmprogramming.com/pure-functions.html and a big reason the procedure composition of actions, reasoning about actions is "easy"... has a declarative and constrained syntax.

amitschang: I see what you are saying, but I don't quite get how this violates the model. In this case we are asking the user to select a time to pump in the action (given as action input), but instead of an arbitrary time, we are suggesting one of two times that are known to be useful. The selected time is then added to the state. The suggested times are not useful outside this action and do not need to be in state (I think)

evolver/hardware/standard/pump.py
imaitland: ok so the underlying serial command is the entire blob of everything communicating on the channel, there is no native concept of discrete hardware. so if you opt out of the discrete abstraction,e.g. "this vial" or "these vials" e.g. None then you have everything.

imaitland (approved):
LGTM

PR #283, opened 2025-03-26
add | None to hardware outputs

imaitland:
Bug fix, sometimes hardware outputs None, since this violates the extant pydantic model it causes issues.


amitschang (approved):
Nice!

PR #287, opened 2025-04-03
Iainland/poly fit none guard

imaitland:
when running evolver locally i have transformer config (that is valid)

```
          output_transformer:
            '0':
              classinfo: evolver.calibration.standard.polyfit.LinearTransformer
              config:
                created: '2025-04-03T11:22:28.839344'
                degree: 1
                dir: calibration_files
                expire: P168D
                name: null
                parameters:
                - 0.6149999999999997
                - 0.024599999999999997
            '1':
              classinfo: evolver.calibration.standard.polyfit.LinearTransformer
              config:
                created: '2025-04-03T11:22:28.813278'
                degree: 1
                dir: calibration_files
                expire: P168D
                name: LinearTransformer
                parameters: null
            '2':
              classinfo: evolver.calibration.standard.polyfit.LinearTransformer
              config:
                created: '2025-04-03T11:22:28.813295'
                degree: 1
                dir: calibration_files
                expire: P168D
                name: LinearTransformer
                parameters: null
```

This is causing errors in the logs because parameters are None but when app is initialized, a transform is attempted against null / None parameters on vials 1 and 2...

worth noting this change (maybe) revealed, un masked another warning, seemingly valid imo, for the vial that has params and can therefore be fitted...

```
dev: commands[0]> python -m evolver.app.main
INFO:     Started server process [38158]
INFO:     Waiting for application startup.
/Users/iainland/Projects/evolver/evolver-ng/.tox/dev/lib/python3.11/site-packages/numpy/polynomial/polynomial.py:1436: RankWarning: The fit may be poorly conditioned
  return pu._fit(polyvander, x, y, deg, rcond, full, w)
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:8080 (Press CTRL+C to quit)
```

Basically the warning is from numpy that the fit on vial 0 is sparse (makes sense since only 2 points...)... 
```
RankWarning: The fit may be poorly conditioned
  return pu._fit(polyvander, x, y, deg, rcond, full, w)
```


evolver/hardware/standard/temperature.py
```diff
@@ -50,8 +50,14 @@ def _do_serial(self, from_proposal=False):
...
-            raw = int(self._transform("input_transformer", "convert_from", input.temperature, vial))
-            data[vial] = str(raw).encode()
+            transformed_value = self._transform("input_transformer", "convert_from", input.temperature, vial)
+            # Handle None results from transformation to prevent int(None) error
+            if transformed_value is not None:
+                raw = int(transformed_value)
+                data[vial] = str(raw).encode()
+            else:
+                # Default to HEAT_OFF if transformation fails
+                data[vial] = self.HEAT_OFF
```
imaitland, line 60: Without this get, following on app start up:

```
INFO:     Waiting for application startup.
/Users/iainland/Projects/evolver/evolver-ng/.tox/dev/lib/python3.11/site-packages/numpy/polynomial/polynomial.py:1436: RankWarning: The fit may be poorly conditioned
  return pu._fit(polyvander, x, y, deg, rcond, full, w)
  ```

amitschang:
This may be a case where we do want the errors in the log. I think the exception fallback logic in the transform function gives the same result, and the errors give a good hint what is the reason behind a None result in the transformed value...

imaitland:
It's masking other warnings though...
And also if a config generates errors shouldn't that be an invalid config?
Maybe we recharacterize them as warnings?

imaitland:
I'd also say generally it's confusing developing against a service with noisy logs, just in the terminal, its not clear what is something critical or ignorable... this was my motivation for this change... i think it is also underlying some of the issues Ezira had last week...where there was a real bug but it was getting lost or masked by other errors e.g. "serial connection failed" etc... we should be able to expect that a valid config does not generate errors in the logs i think...

amitschang:
True indeed. The fact that this is valid is may be the real issue. Does it cause a cascade of issues if the logic rather requires parameters, instead of allowing None?

I wonder if this might be related to the fact that we automatically build an object from config descriptor, so that to specify what class we want to use for transformation in the calibrator - but prior to calibration, and thus parameters - we would have to enable building it without parameters. If so, then this seems a sensible thing to do

amitschang (approved):

imaitland:
yeah its at the "automatically build" step, it's when the evolver starts up the first time.

PR #290, opened 2025-04-04
get resume working

imaitland:
when resuming a calibration procedure, the path the procedure is loaded from using the base config's load method must include the dir for procedure_files.

### DIFF
--- a/evolver/calibration/standard/calibrators/temperature.py
+++ b/evolver/calibration/standard/calibrators/temperature.py
@@ -1,10 +1,7 @@
-from pathlib import Path
-
 from evolver.calibration.action import DisplayInstructionAction
 from evolver.calibration.interface import (
     CalibrationStateModel,
     IndependentVialBasedCalibrator,
-    root_calibrator_file_storage_path,
 )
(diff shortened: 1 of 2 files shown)


evolver/calibration/standard/calibrators/temperature.py
```diff
@@ -26,7 +32,17 @@ def create_calibration_procedure(
...
-        procedure_state = CalibrationStateModel.load(self.procedure_file) if resume else None
+        procedure_state = None
+        if resume and self.procedure_file:
+            # Check if procedure_file is an absolute path
+            if Path(self.procedure_file).is_absolute():
+                procedure_file_path = self.procedure_file
+            else:
+                # Use self.dir if defined, otherwise use the default storage path
+                procedure_dir = getattr(self, "dir", root_calibrator_file_storage_path())
+                procedure_file_path = procedure_dir / self.procedure_file
```
amitschang, line 43: feels a bit like this logic belongs in the `load` method of `CalibrationStateModel`, or rather even its relevent base (I think the `Transformer.Config`.

amitschang (approved):
Nice, thanks!

Issue #247, opened 2025-01-23
Bad classinfo creates Config object instead of failing


amitschang:
In debugging some new code that uses evolver `BaseConfig` with `ConfigDescriptors` as a type within the model, when passing in a `classinfo` that has a typo (and thus does not reference a class that can be imported), the system replaces the object with a plain `Config` object instead of failing.

Ex:

```py
class Experiment(BaseConfig):
    controllers: list[ConfigDescriptor | Controller]
```

and putting in:
```py
{"controllers": [ {"classinfo": "typo", "config": {}} ]}
```
  
caused the error

```
s]
                                         ^^^^^
  File "/Users/arik/ws/ssec/evolver-ng/.tox/dev/lib/python3.11/site-packages/pydantic/main.py", line 811, in __getattr__
    raise AttributeError(f'{type(self).__name__!r} object has no attribute {item!r}')
AttributeError: 'Config' object has no attribute 'run'
```

Showing that what is passed through ends up as a `Config` object. When the typo is corrected, the application runs as expected.

**NB**: This seems to only happen if the type hint is `ConfigDescriptor | Controller`, if it is instead simply `ConfigDescriptor` it winds up erroring about the module.

imaitland:
Any `config: null` will cause this issue.

e.g.

<img width="765" alt="Image" src="https://github.com/user-attachments/assets/a1f85f57-b454-4c75-b383-c5dfe45c881e" />

imaitland:
```
config: {}
```
also ok...

imaitland:
<img width="584" alt="Image" src="https://github.com/user-attachments/assets/d192d275-c65a-47c4-9bf7-0240f47e7b74" />

jamienoss:
@amitschang & @imaitland this is a similar, if not duplicate, issue of #213.

ezirayw:
went through #213 and #216 and saw that potentially putting a ceiling on the pydantic version helped fix problems. not sure if its a bandaid for now but ideally would want to prioritize solutions that work off of the current pydantic version to avoid future code-breaking issues

amitschang:
I think I found some leads here. There are two things going on:

1. For union types, pydantic is trying to validate types and return the first one that properly validates, so when ConfigDescriptor fails (e.g. due to bad import, which in this case is a validation error) it moves on to the other kind (e.g. `Controller`)
1. When it tries to validate a type that inherits from `BaseInterface`, the schema it gets back:
[...]

PR #293, opened 2025-04-10
allow null config in yml

imaitland:
allows nullable configs, per Ezira's requirement, fix made with input and discussion with @jamienoss.

It's a bit of an oddball in that Pydantic's syntax is a little confusing, like surely the state before the fix "means" in spirit what this fix is now explicitly stating... either that or we're not useing Pydantic optional & defaults correctly.

e.g. 

```
hardware:
  test:
    classinfo: evolver.hardware.demo.NoOpSensorDriver
    config:
      calibrator:
        classinfo: evolver.calibration.standard.calibrators.temperature.TemperatureCalibrator
        config: null
```

refs:

https://github.com/ssec-jhu/evolver-ng/issues/247
https://github.com/ssec-jhu/evolver-ng/pull/246
https://github.com/ssec-jhu/evolver-ng/issues/213
https://github.com/ssec-jhu/evolver-ng/pull/216

^ Evidently there is a more gnarly issue at play here, but this addresses the requirement.

### DIFF
--- a/evolver/base.py
+++ b/evolver/base.py
@@ -155,7 +155,7 @@ class BaseConfig(_BaseConfig):
 
 class ConfigDescriptor(_BaseModel):
     classinfo: ImportString
-    config: dict = {}
+    config: dict | None = {}
 
     @pydantic.field_validator("classinfo", mode="after")
     @classmethod


evolver/base.py
```diff
@@ -155,7 +155,7 @@ class BaseConfig(_BaseConfig):
 
 class ConfigDescriptor(_BaseModel):
     classinfo: ImportString
-    config: dict = {}
+    config: dict | None = {}
```
jamienoss, line 158: 👍

amitschang (commented):
Seems like a good idea to me. A None value should basically be synonymous with an empty dict, so that I think we would need to resolve using None in create. This would fail in the case of None:

https://github.com/ssec-jhu/evolver-ng/blob/main/evolver/base.py#L215

So config should be appropriately initialized in that function

evolver/base.py
```diff
@@ -119,21 +119,22 @@ def model_validate(cls, obj, *, strict=None, from_attributes=None, context=None)
                 return cls.model_validate_json(obj, strict=strict, context=context)
             else:
                 # Cool, it was a ConfigDescriptor so validate from that. Remember ``ConfigDescriptor.config`` is just a
-                # dict.
+                # dict or None.
+                config_dict = {} if descriptor.config is None else descriptor.config
                 return super().model_validate(
-                    descriptor.config, strict=strict, from_attributes=from_attributes, context=context
+                    config_dict, strict=strict, from_attributes=from_attributes, context=context
```
amitschang, line 125: It seems to me that we can let the classes validate the input, and a None here would be allowed as per the new config type hint. I think the only place we need to handle the none is in construction...

evolver/tests/test_base.py
```diff
@@ -255,10 +255,21 @@ def test_create(self, mock_descriptor):
...
         assert obj.a == 2
         assert obj.b == 3
 
+        # Test with explicit empty dict config
+        obj = evolver.base.ConfigDescriptor(classinfo="evolver.tests.test_base.ConcreteInterface", config={}).create()
+        assert obj.a == 2
+        assert obj.b == 3
+
+        # Test with None config
+        obj = evolver.base.ConfigDescriptor(classinfo="evolver.tests.test_base.ConcreteInterface", config=None).create()
```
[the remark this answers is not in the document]
amitschang: exactly!

evolver/base.py
```diff
@@ -157,6 +157,11 @@ class ConfigDescriptor(_BaseModel):
     classinfo: ImportString
     config: dict = {}
 
+    @pydantic.field_validator("config", mode="before")
```
imaitland, line 160: important to note this is a "before" validator.

evolver/base.py
```diff
@@ -157,6 +157,11 @@ class ConfigDescriptor(_BaseModel):
...
+    @classmethod
+    def validate_config(cls, v):
+        return {} if v is None else v
```
jamienoss, line 163: Bingo! Good job dude!

amitschang (approved):
this looks good, thanks!

ezirayw:
when testing the new commit https://github.com/ssec-jhu/evolver-ng/commit/8ffbe6b69c8523a163f46e4b6d47c746a543ed89 I still see errors when trying to initiate a calibration procedure from a null config attribute.  Tested by upgrading the backend using `pip install --upgrade git+https://github.com/ssec-jhu/evolver-ng` and checking out https://github.com/ssec-jhu/evolver-ui/tree/iainland/apply-button?tab=readme-ov-file on the frontend 

here is my config file for reference:
image

here is the displayed error in the backend:
image

I also see that a calibrator was created by looking at the configuration tab in the UI so perhaps the error is somewhere in the procedure code (as noted also by the backend error):
image

amitschang:
@ezirayw , looks like that is fixed based on our last debug session but the fix didn't get merged yet. Sorry for the confusion! I will go ahead and merge that PR in now 😄

PR #294, opened 2025-04-11
stick to configdescriptor for configurable objects in config

amitschang:
Unsure why the test is failing... Somehow the driver is complaining about not being json serializable

### DIFF
--- a/evolver/calibration/demo.py
+++ b/evolver/calibration/demo.py
@@ -22,8 +22,8 @@ class NoOpTransformer(Transformer):
 
 class NoOpCalibrator(Calibrator):
     class Config(Calibrator.Config):
-        input_transformer: ConfigDescriptor | Transformer | None = Field(default_factory=NoOpTransformer)
-        output_transformer: ConfigDescriptor | Transformer | None = Field(default_factory=NoOpTransformer)
+        input_transformer: ConfigDescriptor | None = Field(default_factory=NoOpTransformer)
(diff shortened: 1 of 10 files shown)


jamienoss:
Whilst removing these type unions would be the "simplest" way to address #247 - it drastically changes the package. Users/developers will no longer be able to pass around instances as everything will need to be a ``ConfigDescriptor``. The intended functionality is beneficial for developers writing new procedures and classes, e.g., interactively in a notebook. Another clear example are the tests, I believe some of them rely on this functionality.

FYI this PR would decrease the duplicity of this package being both a functional framework and web API, to mostly just that of a web API.

amitschang:
I don't think this change actually makes it so that users/developers can no longer pass instances. All of the other types removed from unions here are derivative of BaseInterface, and this seems to evaluate as `ConfigDescriptor`. For example, with this change, this works just as before:

```py
Evolver.Config(hardware={"test": NoOpSensorDriver()}, history=InMemoryHistoryServer())
```

And validation is happening: replacing either of those objects with strings fails validation because string type is not allowed. 

I think the test failure is actually more indicative of some symmetry issue, possibly with nesting or maybe specifically related to calibration. I can equally save the above config to disk. 

The thing this does do is make the deserialization more predictable in that it does not fallback to the config type object. This isn't contractual though, which means that a user developing a component can add a union and arrive at the same trouble from #247 - we'd only be able to advise against that

PR #295, opened 2025-04-15
making all things deserialize as configdesc, iso config

amitschang:


imaitland:
general question, did you try and upgrade pydantic and run the tests? There's one that currently fails on any pydantic upgrade... that this PR may resolve...

amitschang:
@imaitland , good point. I ran tests after an upgrade to 2.11.3 (latest as of now) and looks like there is only a trivial json_schema representation error (I think it is trivial). I'll work on that and update shortly if I can

amitschang:
it was trivial, the json schema added an "additionalProperties" field set to true - which is already the default in json schema (see https://json-schema.org/understanding-json-schema/reference/object#additionalproperties)

imaitland:
Hell yeah nice! Congrats 🍾

imaitland (approved):
LGTM!

imaitland:
Bit of a side question, and for the sake of my own understanding, Jamie mentioned in another comment that one of the nice things about the union types with ConfigDescriptor is that it allows using the Evolver framework in python environments, e.g. a jupyter notebook, in a fashion similar to how you'd use an SDK.

when he said that this is what i imagined:

e.g.
```
myEvolver = new Evolver()
myHardware = new Hardware()
myCalibrationProcedure = new CalibrationProcedure()
myHardware.calibrator.calibration_procedure = myCalibrationProcedure
myEvolver.hardware[0] = myHardware

...etc
```

is this understanding correct? if so presumably this kind of functionality remains after the changes in this PR?

amitschang:
@imaitland, I believe that understanding is correct. And yes, there is no changes here wrt that, one can still:

```py
Evolver.Config(hardware = {'od': ODSensor(addr='od90', integrations=100)})
```

or 

```py
Evolver(hardware = {'od': ODSensor(addr='od90', integrations=100)})
```

PR #298, opened 2025-04-25
Iainland/procedure perf bug

imaitland:
Procedure history is a LIFO stack that duplicated itself within itself... causing perf issues.

Fix is to only keep comprehensive history on the head of the LIFO.

Added a test - 100 actions are now dispatched and undone in `0.1s`!


.gitignore
```diff
@@ -175,3 +175,4 @@ openapi.json
 **/calibration_files/
 **/experiment_files/
 **/evolver_snapshot.yml
+**/app/tests/**evolver.yml
```
imaitland, line 178: running tests 1 by 1 in pycharm sometimes generates a config in this dir and i accidentally commit it.

evolver/app/tests/test_calibration.py
```diff
@@ -364,15 +364,14 @@ def test_get_calibration_data(tmp_path):
             {"completed_actions": [], "history": [], "measured": {}, "started": True},
             {
                 "completed_actions": ["measure_vial_0_temperature"],
-                "history": [{"completed_actions": [], "history": [], "measured": {}, "started": True}],
+                "history": [],
```
imaitland, line 367: no recursive history

Issue #300, opened 2025-04-30
[calibration][temperature] temp calibration should take median of 3 reads, take measurements at 2 temps


amitschang:
* Each time we read the raw value from the hardware, we should take median of 3 values
* We should take readings for all vials in several waves:
  * At room temperature (heater off)
  * At some higher temperature (heater on to X - X is some serial value that is reasonably pre-determined)

PR #302, opened 2025-04-30
add 3 raw readings and median, do 3 temp readings

imaitland:
refs: https://github.com/ssec-jhu/evolver-ng/issues/300

### DIFF
--- a/evolver/calibration/standard/actions/temperature.py
+++ b/evolver/calibration/standard/actions/temperature.py
@@ -26,14 +26,14 @@ class RawValueAction(CalibrationAction):
     class FormModel(BaseModel):
         pass
 
-    def __init__(self, vial_idx: int, *args, **kwargs):
+    def __init__(self, vial_idx: int, num_readings: int = 3, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.vial_idx = vial_idx
+        self.num_readings = num_readings
 
     def execute(self, state: CalibrationStateModel, payload: Optional[FormModel] = None):
         all_raw_readings = []
-        NUM_READINGS = 3
-        for i in range(NUM_READINGS):
+        for i in range(self.num_readings):
             raw_reading = self.hardware.read()[self.vial_idx].raw
             all_raw_readings.append(raw_reading)
         median_raw_reading = np.median(all_raw_readings)
--- a/evolver/calibration/standard/actions/temperature.py
+++ b/evolver/calibration/standard/actions/temperature.py
@@ -1,6 +1,7 @@
 from collections import defaultdict
 from typing import Optional
 
+import numpy as np
 from pydantic import BaseModel, Field
 
 from evolver.calibration.action import CalibrationAction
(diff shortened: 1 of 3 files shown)


evolver/calibration/standard/actions/temperature.py
```diff
@@ -25,14 +26,19 @@ class RawValueAction(CalibrationAction):
...
         self.vial_idx = vial_idx
+        self.num_readings = num_readings
 
     def execute(self, state: CalibrationStateModel, payload: Optional[FormModel] = None):
-        sensor_value = self.hardware.read()[self.vial_idx].raw
+        all_raw_readings = []
+        for i in range(self.num_readings):
+            raw_reading = self.hardware.read()[self.vial_idx].raw
+            all_raw_readings.append(raw_reading)
+        median_raw_reading = np.median(all_raw_readings)
```
amitschang, line 39: a somewhat more python way to do this would be (using list comprehension iso append):

```py
readings = [self.hardware.read()[self.vial_idx].raw for i in range(self.num_readings)]
median_raw = np.median(readings)
```

evolver/calibration/standard/calibrators/temperature.py
```diff
@@ -58,22 +61,30 @@ def create_calibration_procedure(
             )
         )
 
-        for vial in self.vials:
+        for i in range(self.num_temp_readings):
             calibration_procedure.add_action(
-                ReferenceValueAction(
+                DisplayInstructionAction(
+                    description=f"Beginning vial sweep {i} of {self.num_temp_readings} set temperature and wait 25 mins for global equilibrium",
```
amitschang, line 67: could we have an action that sets the appropriate temperature (or temp raw val) between each sweep?

amitschang: BTW, here is the relevant code (I think) from the existing procedure: https://github.com/FYNCH-BIO/evolver-electron/blob/75f01fea44983be0260f6d6697440f1de0c4dfa3/app/components/TempCalibrate.js#L323

Looks like there is a pre-configured range and then each step bumps an equal portion of the range. So for a range of 1000, there would be 0, 500, 1000 for 3 steps. It seems to indicate that it is 0.02C per unit, so 500 is +10C. There is some accounting for room temperature that I'm not totally sure about.

imaitland: probably a good question for Ezira.

evolver/calibration/standard/actions/temperature.py
imaitland: will fix

evolver/calibration/standard/calibrators/temperature.py
imaitland: I asked Ezira on slack. I am going to have to confine my contributions to the framework side, as distinct from specific implementation details for now. The framework will certainly support the procedure doing this.

amitschang (approved):

PR #305, opened 2025-05-06
Add set temperature action to calibration waves

amitschang:
This:
* For fist wave: Turns off the heater 
* For all other waves: Sets the heater to a raw adc value proportional to number of steps within a configurable range, relative to the read-in room temperature settings (taken during the first wave). For example:
  * if raw reading taken at first wave for vial 0 is `2048`, range `(heater_boundary_low, heater_boundary_high)` is `(0, 1000)`, and `num_temp_readings` is `3`. The second wave for vial 0 will be set at `1548` and the second will be `1048`
  * For the temp sensor this corresponds to approximately +10C and +20C for three waves (sensor ~0.02C/unit)

It also fixes the temp driver to allow raw to be set in input

This was modeled after https://github.com/FYNCH-BIO/evolver-electron/blob/master/app/components/TempCalibrate.js, or at least my reading of it

### DIFF
--- a/evolver/calibration/standard/actions/temperature.py
+++ b/evolver/calibration/standard/actions/temperature.py
@@ -46,11 +46,11 @@ class AdjustHeaterAction(CalibrationAction):
         self.raw_adjustment = raw_adjustment
 
     def execute(self, state: CalibrationStateModel, payload=None):
-        current_readings = self.hardware.get()
         for vial in self.vials:
-            # current readings could come average of historical readings using
-            # history server.
-            raw = int(current_readings[vial].raw + self.raw_adjustment)
+            # To do the adjustment action, we have to have read the room temp
+            # reading, and it has to have been the first reading measured.
+            room_temp_raw = state.measured[vial]["raw"][0]
+            raw = int(room_temp_raw + self.raw_adjustment)
             self.hardware.set(vial=vial, temperature=None, raw=raw)
         self.hardware.commit()
         return state
--- a/evolver/calibration/standard/actions/temperature.py
+++ b/evolver/calibration/standard/actions/temperature.py
@@ -39,6 +39,35 @@ class RawValueAction(CalibrationAction):
         return state
 
 
+class AdjustHeaterAction(CalibrationAction):
+    def __init__(self, vials: list[int], raw_adjustment: int, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.vials = vials
+        self.raw_adjustment = raw_adjustment
+
+    def execute(self, state: CalibrationStateModel, payload=None):
+        current_readings = self.hardware.get()
+        for vial in self.vials:
+            # current readings could come average of historical readings using
+            # history server.
+            raw = int(current_readings[vial].raw + self.raw_adjustment)
+            self.hardware.set(vial=vial, temperature=None, raw=raw)
+        self.hardware.commit()
+        return state
+
+
+class HeaterOffAction(CalibrationAction):
+    def __init__(self, vials: list[int], *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.vials = vials
+
+    def execute(self, state: CalibrationStateModel, payload=None):
+        for vial in self.vials:
+            self.hardware.set(vial=vial, temperature=None, raw=None)
+        self.hardware.commit()
+        return state
+
+
 class CalculateFitAction(CalibrationAction):
     class FormModel(BaseModel):
         pass
--- a/evolver/calibration/standard/actions/temperature.py
+++ b/evolver/calibration/standard/actions/temperature.py
@@ -39,7 +39,7 @@ class RawValueAction(CalibrationAction):
         return state
 
 
-class AdjustHeaterAction(CalibrationAction):
+class AllVialsAdjustHeaterAction(CalibrationAction):
     def __init__(self, vials: list[int], raw_adjustment: int, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.vials = vials
@@ -47,16 +47,13 @@ class AdjustHeaterAction(CalibrationAction):
 
     def execute(self, state: CalibrationStateModel, payload=None):
         for vial in self.vials:
-            # To do the adjustment action, we have to have read the room temp
-            # reading, and it has to have been the first reading measured.
-            room_temp_raw = state.measured[vial]["raw"][0]
-            raw = int(room_temp_raw + self.raw_adjustment)
+            raw = int(state.roomTempRawValues[vial] + self.raw_adjustment)
             self.hardware.set(vial=vial, temperature=None, raw=raw)
         self.hardware.commit()
         return state
 
 
-class HeaterOffAction(CalibrationAction):
+class AllVialsHeaterOffAction(CalibrationAction):
     def __init__(self, vials: list[int], *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.vials = vials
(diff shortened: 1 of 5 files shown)


imaitland:
good one for Ezira to review imo.

evolver/calibration/standard/calibrators/temperature.py
```diff
@@ -53,18 +58,39 @@ def create_calibration_procedure(
...
+                )
+                approx_delta = -1 * self.heater_slope_init * adjustment  # This is just for display purposes
+                calibration_procedure.add_action(
+                    AdjustHeaterAction(
+                        name=f"vial_sweep_{i}_adjust_heaters",
+                        vials=self.vials,
+                        raw_adjustment=adjustment,
+                        description=(
+                            f"Begin sweep for temperature wave {i + 1}/{self.num_temp_readings}."
+                            f"Will set heaters to approximately room temp +{approx_delta:0.2f} degrees"
```
imaitland, line 85: not sure if this is celcius of farenheit.

evolver/calibration/standard/calibrators/temperature.py
```diff
@@ -53,18 +58,39 @@ def create_calibration_procedure(
...
+                        ),
+                        hardware=selected_hardware,
+                    )
+                )
+
             calibration_procedure.add_action(
                 DisplayInstructionAction(
-                    description=f"Beginning vial sweep {i} of {self.num_temp_readings} set temperature and wait 25 mins for global equilibrium",
+                    description="Wait global equilibrium. This may take several hours. Proceed when reached.",
```
imaitland, line 93: "wait for global equilibrium..."

evolver/hardware/standard/temperature.py
```diff
@@ -23,7 +23,8 @@ class Temperature(SensorDriver, EffectorDriver):
     stands in for off (HEAT_OFF property of this class).
     """
 
-    HEAT_OFF = b"4095"
+    HEAT_OFF_RAW = 4095
+    HEAT_OFF = str(HEAT_OFF_RAW).encode()
```
imaitland, line 27: Comment for what this is, is it the raw value that turns the heater off? or a limit beyond which the heater will shut down?

amitschang:
@ezirayw Could you take a look and review? I can't get you in the review list, maybe cause you need to be explicitly on the repo, so I sent invite.

evolver/calibration/standard/actions/temperature.py
```diff
@@ -39,6 +39,35 @@ def execute(self, state: CalibrationStateModel, payload: Optional[FormModel] = N
...
+class AdjustHeaterAction(CalibrationAction):
+    def __init__(self, vials: list[int], raw_adjustment: int, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.vials = vials
+        self.raw_adjustment = raw_adjustment
+
+    def execute(self, state: CalibrationStateModel, payload=None):
+        for vial in self.vials:
+            # To do the adjustment action, we have to have read the room temp
+            # reading, and it has to have been the first reading measured.
```
imaitland, line 51: I think this would be better as an explicit attribute in the procedure state, e.g. an action "zeroHeaterAction" that updates an explicit attribute (e.g. "zeroedHeaterRawValue" in the state. right now this behavior is implicit not explicit in the procedure...

amitschang: I like the updating explicit attribute. For the action, we can do this as one action for all vials I guess, then it wouldn't add much overhead, I guess - let me try that

evolver/calibration/standard/actions/temperature.py
```diff
@@ -39,6 +39,35 @@ def execute(self, state: CalibrationStateModel, payload: Optional[FormModel] = N
...
+            room_temp_raw = state.measured[vial]["raw"][0]
+            raw = int(room_temp_raw + self.raw_adjustment)
+            self.hardware.set(vial=vial, temperature=None, raw=raw)
+        self.hardware.commit()
+        return state
+
+
+class HeaterOffAction(CalibrationAction):
```
imaitland, line 59: "AllVialsHeaterOffAction"

imaitland: or "AllVialHeatersOffAction"

evolver/calibration/standard/actions/temperature.py
```diff
@@ -39,6 +39,35 @@ def execute(self, state: CalibrationStateModel, payload: Optional[FormModel] = N
         return state
 
 
+class AdjustHeaterAction(CalibrationAction):
```
imaitland, line 42: AdjustVialHeatersAction

evolver/calibration/standard/calibrators/temperature.py
```diff
@@ -25,6 +27,9 @@ class Config(IndependentVialBasedCalibrator.Config):
         num_temp_readings: int = Field(
             3, description="Number of times reference temperature readings are taken from each vial."
         )
+        heater_boundary_low: int = 0
+        heater_boundary_high: int = 1000
+        heater_slope_init: float = 0.02
```
imaitland, line 32: nice to have a description on these.

evolver/calibration/standard/calibrators/temperature.py
imaitland: ah nvm `f`

evolver/hardware/standard/temperature.py
```diff
@@ -35,7 +36,8 @@ class Output(SerialDeviceOutputBase, SensorDriver.Output):
         temperature: float | None = Field(None, description="Sensor temperature in degrees Celsius")
 
     class Input(EffectorDriver.Input):
-        temperature: float = Field(description="Target temperature in degrees Celsius")
+        temperature: float | None = Field(None, description="Target temperature in degrees Celsius")
```
imaitland, line 39: see earlier comment about units, are we in C or F?

evolver/hardware/standard/temperature.py
```diff
@@ -50,7 +52,10 @@ def _do_serial(self, from_proposal=False):
             inputs.update({k: v for k, v in self.proposal.items() if k in self.vials})
         for vial, input in inputs.items():
             # Calibrate temperature to raw data.
-            raw = int(self._transform("input_transformer", "convert_from", input.temperature, vial))
+            if input.temperature is None:
+                raw = input.raw if input.raw is not None else self.HEAT_OFF_RAW
```
imaitland, line 56: inline comment or maybe even a docstring calling out that if input.temperature is None and input.raw is None, the "off" signal will be sent (as i understand it). since this is implicit behavior ...

evolver/hardware/standard/temperature.py
ezirayw: This is effectively an input to the heater to turn off since we setting the set point to the lower limit of the PID controller. The sensor will still collect and send out raw temp data (i.e. room temp) so they are de-coupled. A PID controller is used to coordinate the two at the Arduino level (input set points from the user & real-time sensor data dictate what power level the heater needs to be at to maintain the desired set point)

evolver/calibration/standard/actions/temperature.py
ezirayw: AllVialsHeaterOffAction makes more sense to me here

evolver/hardware/standard/temperature.py
ezirayw: C

evolver/calibration/standard/calibrators/temperature.py
ezirayw: yeah, I think having inline units should suffice

evolver/hardware/standard/temperature.py
amitschang: added to docstring of input class

evolver/calibration/standard/actions/temperature.py
amitschang: I did this, and it seems much better indeed!

evolver/calibration/standard/calibrators/temperature.py
amitschang: added descriptions via Field values

evolver/hardware/standard/temperature.py
amitschang: I added some comment strings for clarity. I think we match the device behavior - but if not, that should be resolved elsewhere, the change here is primarily for supported direct setting raw, which was missed first pass

evolver/calibration/standard/calibrators/temperature.py
```diff
@@ -53,22 +61,54 @@ def create_calibration_procedure(
...
             calibration_procedure.add_action(
                 DisplayInstructionAction(
-                    description=f"Beginning vial sweep {i} of {self.num_temp_readings} set temperature and wait 25 mins for global equilibrium",
+                    description="Wait for global equilibrium. This may take several hours. Proceed when reached.",
                     name=f"vial_sweep_{i}_wait_for_equilibrium_instruction",
                     hardware=selected_hardware,
                 )
             )
+
+            if i == 0:
```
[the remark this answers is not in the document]
amitschang: this reduces a bit more repetition than moving it up there, since we still have to do the waiting step that is directly above this

imaitland (approved):
LGTM

PR #306, opened 2025-05-20
run loop in thread in main application

amitschang:
Since the device manager loop activity is not itself async and has some blocking sections, we run within a thread instead of coroutine. This also implements the loop wait interval to an event with timeout in order to support on-demain re-evaluation of the loop during update.

### DIFF
--- a/evolver/app/main.py
+++ b/evolver/app/main.py
@@ -42,14 +42,14 @@ async def lifespan(app: FastAPI):
 
 app = FastAPI(lifespan=lifespan, default_response_class=ORJSONResponse)
 app.state.evolver = None
-app.state.trigger = threading.Event()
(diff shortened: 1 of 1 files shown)


evolver/app/main.py
```diff
@@ -42,6 +42,14 @@ async def lifespan(app: FastAPI):
 
 app = FastAPI(lifespan=lifespan, default_response_class=ORJSONResponse)
 app.state.evolver = None
+app.state.loop_trigger = threading.Event()  # enables on-demand re-executon of the loop
```
imaitland, line 45: is there an example of the re-execution being triggered?

imaitland: nvm presuming: `app.state.loop_trigger.set()` in the update method is it.

imaitland (approved):
lgtm

PR #308, opened 2025-05-27
Docs update

amitschang:
This PR addresses the below feedback (from slack) in a way I think fits with the package:

> * use "eVOLVER" spelling to maintain consistency (outside of code blocks)

✅ also added a note with disambiguation of use through the docs

> * the config section is still blank but when its populated i think it would useful to show an example of a valid config and the values each attribute expects
> * a page/section dedicated to explaining the core data validation models

^ The first is addressed in the config section pointing to its generic nature and the json schema endpoint/methods, the config class pattern, and the relevant definitions for the application. The second one is addressed in the same config section pointing to its pydantic base and example field definitions.

> * when describing how to develop custom eVOLVER components, should also include how to explicitly integrate into the system. can be as simple as "ssh x into y directory" but i having it documented ensures that people are putting custom things in the right places so that when the config is edited to incorporate there arent any issues

^ added in an installation page under the development section to discuss how to get custom components available on the system.

> * in the installation page include a page for the UI (most people will be interfacing with the config through the UI so i think having a basic installation guide for it is useful here)

^ Included a section for the UI installation and service definition on the rPi - with a pointer to the UI repo

### DIFF
--- a/docs/source/concepts.rst
+++ b/docs/source/concepts.rst
@@ -1,7 +1,7 @@
-Evolver concepts
+eVOLVER concepts
 ================
 
-The evolver system is made up of several configurable components tied together
+The eVOLVER system is made up of several configurable components tied together
 by a manager component that orchestrates the application control loop and
 exposure to the web interface. Components can be grouped into two high level
 categories:
@@ -42,7 +42,7 @@ categories:
 Configuration and ConfigDescriptors
 -----------------------------------
 
-All components in the evolver can be represented by a so-called
+All components in the eVOLVER can be represented by a so-called
 :py:class:`ConfigDescriptor<evolver.config.ConfigDescriptor>`. This is the key
 component which translates from a static serialized configuration (e.g. that
 which can be represented in a yaml file) to the in-memory instantiated objects
@@ -62,7 +62,7 @@ particular hardware might look like:
           addr: "od90"
           integrations: 500
 
-The object under ``hardware.temp`` is recognized in the evolver system as a
+The object under ``hardware.temp`` is recognized in the eVOLVER system as a
 ``ConfigDescriptor``, where the object to create is an
 ``evolver.hardware.standard.od_sensor.ODSensor`` and the config to pass in will
 be an evaluated version of its ``Config``. To clarify the point, consider the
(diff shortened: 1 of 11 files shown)


docs/source/development/config.rst
```diff
@@ -1,4 +1,171 @@
...
+            "required": [
+                "param_required"
+            ],
+            "title": "Config",
+            "type": "object"
+        }
+    }
+
+
+which contains enough information to generate a UI form for the component's
```
imaitland, line 122: not really material to this PR but noting here for comepleteness - the name attribute with "anyOf" that includes null, and defaults to null, sadly isn't sufficient to generate a ui (html) form input element. If it comes in as "null" the html form will not render an input element since null has no input... it knows it could be a string.

the better way to do this would be to either default to an empty string...

docs/source/concepts.rst
```diff
@@ -1,7 +1,7 @@
-Evolver concepts
+eVOLVER concepts
```
imaitland, line 1: also not material to this PR, and i know it's something the PI has asked for / insisted upon -  but this "branding" is on so many levels cursed... 

i am speaking here as an aficionado of bauhaus design principles https://www.bauhaus-bookshelf.org/bauhaus_writing_in_small_letters_lower_case_only.html

docs/source/development/config.rst
imaitland: On this note, it's also worth adding, that the FQCN classinfo attribute, from which the html form elements are derived cannot be rendered or validated using the auto-generation mechanisms the UI might use. In this way, the caveat to any promise of "auto-generated ui form elements based on config attributes" depends on the FQCN classinfo being correct when it is added or changed.

docs/source/installation.rst
```diff
@@ -84,4 +87,66 @@ installing the package, and example configurations to simplify setup for
 end-users. See :doc:`development/index` for more information on creating
 extensions.
 
+Web UI
+------
```
imaitland, line 91: I *think* it's worth referring to the ui readme here and not reproducing it, in case they diverge...

imaitland (approved):
LGTM, thanks for doing this. I can definitely put together the docs for the UI, good to know where they live.

docs/source/development/config.rst
amitschang: do you mean for json-schema anyOf is not specified completely? In any case if there are particular issues here we should probably make a ticket and then we can figure out what caveats should be added or if there are reasonable fixes

